PR #32, opened 2014-03-27
Six suggested improvements

jeisfeld:
One year after discussing, finally I managed to access GitHub to make the suggested changes.

Unfortunately, some files are marked by Git as completely changed, although I updated only a few lines - I still did not find out why and when this happened and how I can prevent this. (I guess that it is related to line breaks.)

Suggested Changes (ordered by the five commits)

- ContainsIgnoreCaseValidationMatcher: this is a bug. The change is just a few characters in line 31 - I don't know why Git thinks that I have replaced the whole file.

- TestCase: some suggestion to add variables that can be used in logging.

- RandomStringFunction: Here I just removed one "System.out.println" line - again the file is marked as completely replaced :-(.

- A slight modification in five classes related to Groovy scripts. These changes enabel to have a combination of file based scripts and inline scripts. This is helpful if there are reusable parts that should be applied in multiple Groovy scripts (such as helper methods, making use of context variables), but still it is intended to keep the main Groovy script directly in the test case.

- I included parameter TestContext in the methods of interface TestActionListener. This has no immediate impact, but it is helpful when you want to include more logging information in some custom logger - we use this and the second improvement in order to make the log more readable in case of parallel execution of tests. (For performance reasons, we run always 20 threads in parallel, and the default logging is completely unreadable in such a case - we needed to include in each log line information about the test case which this line refers to.)

- In TestContext, I used ConcurrentHashMap rather than LinkedHashMap to store context variables. LinkedHashMap is not thread safe. If a test case has parallel actions, and one thread adds a variable leading to an increase of capacity of LinkedHashMap, while another thread tries to read variables, this leads sporadically to a "Unknown variable" exception. This is fixed by using the thread safe ConcurrentHashMap.

These changes are included in our Citrus test (based on Citrus 1.3.1). It would be gread to include them in version 1.4, so that we can reduce our customizations (and others can use them as well).
No further changes are planned in the near future ;-).

### DIFF
--- a/modules/citrus-admin/src/main/java/com/consol/citrus/admin/websocket/WebSocketTestListener.java
+++ b/modules/citrus-admin/src/main/java/com/consol/citrus/admin/websocket/WebSocketTestListener.java
@@ -18,7 +18,9 @@ package com.consol.citrus.admin.websocket;
 
 import com.consol.citrus.TestAction;
 import com.consol.citrus.TestCase;
+import com.consol.citrus.context.TestContext;
 import com.consol.citrus.report.*;
+
 import org.json.simple.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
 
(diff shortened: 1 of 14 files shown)


christophd:
Closed as contributions were merged manually, thanks again!

PR #36, opened 2015-02-04
SOAP MTOM attachment support

Stonator:
Bug fixes:
* Support loading of WSDL files and their imports from jar files

Enhancements:
* Support of mtom for attachements (XOP package), new attribute "mtom-enabled" (defaults to false)
* Support mtom xop references and mtom inline data with auto-detection of data type, new attribute "mtom-inline" (defaults to false)
* Binary attachment support (no dynamic data replacement, no string conversion) based on content type
* Improved handling of dynamic content in attachments
* Added "schema-repository" to SendSoapMessageAction to support defining a custom schema repository to resolve mtom attachments
* Added SOAP mtom attachment integration test

Known issue:
* Schema validation for SOAP messages with MTOM attachments (not inline) is not working, because the XOP reference is not valid data. To support schema validation, XOP references need to be resolved and attachments must be inlined before validation. Inline MTOM attachments are validated successfully. An example of this issue is added to the integration tests.

### DIFF
--- a/modules/citrus-core/src/main/java/com/consol/citrus/xml/schema/JarWSDLLocator.java
+++ b/modules/citrus-core/src/main/java/com/consol/citrus/xml/schema/JarWSDLLocator.java
@@ -1,0 +1,70 @@
+package com.consol.citrus.xml.schema;
+
+import java.io.IOException;
+import java.net.URI;
+import javax.wsdl.xml.WSDLLocator;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
+import org.xml.sax.InputSource;
+
+/**
+ * Locates WSDL import sources in Jar files
+ */
+public class JarWSDLLocator implements WSDLLocator {
+
+    private Resource wsdl;
+    private Resource importResource = null;
+
+    public JarWSDLLocator(Resource wsdl) {
+        this.wsdl = wsdl;
+    }
+    
+    @Override
+    public InputSource getBaseInputSource() {
+        try {
+            return new InputSource(wsdl.getInputStream());
+        } catch (IOException e) {
+            return null;
+        }
+    }
+
+    @Override
+    public InputSource getImportInputSource(String parentLocation, String importLocation) {
+        String resolvedImportLocation;
+        URI importURI = URI.create(importLocation);
+        if (importURI.isAbsolute()) {
+            resolvedImportLocation = importLocation;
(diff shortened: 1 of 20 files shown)


christophd:
Thanks a lot, this is what makes open source software so nice! I struggled earlier to get MTOM working. Just a view things that I have changed:
- put attachment encoding type (base64, hex) to attachment itself instead of reading the type from xsd schema. This way Citrus is possible to send invalid encoding in error scenarios.
- refactored soap attachment content handling (static String content, content resource and binary data handler)
- some minor adjustments

Known issue with schema validation is not fixed yet.

Stonator:
Hello Christoph,

As you might remember I have contributed the mtom attachment support for SOAP. In this commit I have indirectly also fixed an issue with resource attachment handling, but unfortunately during merging you reverted this fix and named it optimization. Maybe the fix was not clear in that context.

Problem:
Assume you have a template to upload a file to a SOAP server. The file is given as template parameter variable, and the attachment is a resource using this variable. When you call this twice the second call takes the variable of the first call because it is resolved too early.

For example:

<template name=”upload”>
                <ws:send endpoint=”…”>
<message>
<payload/>
</message>
<ws:attachment content-id="${dpx.uploadfilename}" content-type="application/octet-stream">
<ws:resource file="${param.filename}"/>
</ws:attachment>
</ws:send>
</template>

<testcase>
                Here we call the template to upload file 1
                <call-template name=”upload”>
                                <parameter name=”param.filename” value=”classpath:file1”/>
                <call-template>
                Here we call the template to upload file 2
                <call-template name=”upload”>
                                <parameter name=”param.filename” value=”classpath:file2”/>
                <call-template>

The problem is, that on the first template call in SoapAttachment.java:305 the variable ${param.filename} is resolved by “classpath:file1”. The problem is that the second template call uses the same instance as the first template call and nothing is resolved anymore, meaning that two times the same file is uploaded and file2 is totally ignored.

I fixed this in commit d1316344db2437b83d89a35229673a14e9f8fea3 by creating always a new “resolved” data handler (in my commit SoapAttachment.java:319) whenever resolve is called. I keep the unresolved values to resolve them again on the next call. This ensures that changed variables are always taken into account.
You optimized that out in commit ca8166f6cedde62a736d706fef5e5a7d5cf06f6d and now the bug is here again. Could you please review your changes and fix this? Otherwise I need to fix this again in a custom branch as we really need that feature.

Kind regards,
Reinhard

PR #48, opened 2015-10-27
Fixed XML schema validation of WSDL based schemas. When a WSDL is imp…

vdsrd:
Fixed XML schema validation of WSDL based schemas.

When a WSDL is importing multiple schemas, the current validation loaded the schemas in the wrong order. The order is important for the XmlValidatorFactory.createValidator to create a schema validator. This factory needs to know every type when it is used. Meaning for example, when a WSDL schema imports a schema SA with a type TA and uses it, the schema SA needs to be in the returned resource array "before" the WSDL schema, otherwise building a validator will fail with an exception that type TA is unknown.
This was implemented correct for all imported schemas (first add imported schemas before the parent schema to the resource array), but wrong for the embedded WSDL schema (root schema) which was added on position 0 instead of at the end. Added an integration test to test this behavior.

### DIFF
--- a/modules/citrus-core/src/main/java/com/consol/citrus/xml/schema/WsdlXsdSchema.java
+++ b/modules/citrus-core/src/main/java/com/consol/citrus/xml/schema/WsdlXsdSchema.java
@@ -105,6 +105,7 @@ public class WsdlXsdSchema extends SimpleXsdSchema implements InitializingBean {
         Types types = definition.getTypes();
         List<?> schemaTypes = types.getExtensibilityElements();
         boolean xsdSet = false;
+        Resource firstSchemaInWSDL = null;
         for (Object schemaObject : schemaTypes) {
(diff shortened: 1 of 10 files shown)


vdsrd:
I have reviewed your changes, but you missed the point with the first schema. Because of the initial change you have applied, the WSDL internal schema is now on the last position of the array. Therefore  I have introduced a variable "firstSchemaInWSDL" which is the first internal WSDL schema, that will be used as primary schema reference.

To test this behavior, I have also used by intention two different target namespaces in the WSDL which you have changed as well. My initial WSDL in the integration test is not working anymore with your merged implementation, but it is a valid WSDL which we face in our citrus testing as well. That's the reason why this fix is included.

So could you please use my initial WSDL and schemas in the test, so you will see why I did the fixes in the source.

vdsrd:
Superseded by #54

PR #54, opened 2015-11-23
Fixed XML schema validation, follow up to pull request #48

vdsrd:
Fixed XML schema validation when namespace of WSDL schema does not match namespace of WSDL. Added sample schema and integration test.

Follow up of pull request #48, commit db116dea881de88150fc2f75ca4301e37f3923bc.


christophd:
Thanks a lot for the contribution!

PR #56, opened 2015-12-02
Fix typo "paylaod" --> "payload"

gucce:
I fixed the typo in `paylaod` to `payload`. I am not sure if the exception message in `CitrusRuntimeException` is parsed somewhere, in which case this change could lead to incompatibilities. Otherwise, the change is a very small one.


christophd:
Thanks!

PR #98, opened 2016-05-17
Added 'root-parent-context' to citrus-ws:server for SOAP1.2

bjoerndev:
After some struggle getting `citrus-ws:server` to work with SOAP 1.2 using the documentation [1], I found the trick after debugging the code and reading an example [2] by adding `root-parent-context="true"` to `citrus-ws:server`.
To save this time for others, I added this line to the documentation

[1] http://citrusframework.org/reference/html/index.html#soap-12
[2] https://github.com/christophd/citrus/blob/master/modules/citrus-ws/src/test/resources/citrus-context.xml


bjoerndev:
Thanks for merging, and for providing Citrus :)

PR #138, opened 2016-10-17
Fix unescaped brackets

gucce:
Text elements describing XML tags (enclosed in '<' and '>') are not displayed in HTML since they are recognized as HTML tags. They must be escaped in Markdown.


christophd:
Thank you very much!

PR #172, opened 2016-11-28
simply copy byte array out of input stream

aaschmid:
* do not convert binary data using encoding as this most likely break binary data
* also create realistic test case which actually fails with old code (using citrus.png)

Do not hesitate to ask if you have any questions :-)


christophd:
Thanks a lot!

aaschmid:
No, thank you for that framework :-)

PR #189, opened 2017-01-16
Groovy caching support

vdsrd:
We added Groovy caching support.

The functionality is enabled by default and can be disabled using a system property "cache-groovy-source=false" or set this variable in the citrus.properties file.

It is also possible to disable it per "Groovy Action" definition in a test case by using the 'cacheSource' parameter:
```
<ci:groovy cacheSource="false">
...
</ci:groovy>
```

This change improves the performance significantly and uses less memory.


svettwer:
Hi!

Even if this might be a valuable addition to Citrus, this PR is now nearly two years old and I haven't had the time to update this branch to the current master. Therefore I'll close this PR.

I'm not happy with that at all but I think we've to move on and leave some things undone. :sob: 

BR,
Sven

PR #251, opened 2017-06-06
Test MetaInfo for Java DSL and invoke JIRA API library on Citrus Test Failure

viharnaik:
1.  TestMetaInfo annotation which stores testcase information such as

    description
    requirementID
    scenario
    author

2. CSV report generation from the test output.
- All the test cases can be imported into JIRA through JIRA import plugin from citrus csv file.

CSV report gets generated with the listed information and mapping is

 
@CitrusTest.name - Test Name
@TestMetaInfo.description- Test Description
Place Holder for JIRA import- Test Step
Place Holder for JIRA import- Test Data
Place Holder for JIRA import- Expected Result
UUID- External ID
Current time- Time Stamp
Place Holder for JIRA import- Components
Place Holder for JIRA import- Priority
@TestMetaInfo.author- Author
Place Holder for JIRA import- Labels
@TestMetaInfo.requirementID- Requirement Id
@TestMetaInfo.scenario- Scenario
Test output- Actual Result
Exception if any	Failure Reason

**csv file generated control is similar as html report**

**citrus.csv.report.enabled** : Enables/disables CSV report generation (default= true).
**citrus.csv.report.directory** : Output directory path (default= ${project.build.directory}/test-output/citrus-reports).
**citrus.csv.report.file** : File name for the report file (default= citrus-test-results-dateandtime.csv).

3. On Test Faillure, JIRA API will be invoked which we created for our own organization to create failure report through API calls.


christophd:
One question will users that do not have the failure report API in JIRA be able to use this extension in a profitable way?

svettwer:
Any news here @viharnaik ?

viharnaik:
This only works with JIRA APIs. User without API in JIRA will not be able to use it.

svettwer:
Hi again!

We discussed this PR internally and agreed that this would be a great extension but more as a plug-in instead of a core feature of the framework.
To achieve this, we'll have to design a SPI for citrus. This would enable anybody to add isolated implementations for citrus related features as separate projects. Unfortunately this is nothing we can provide in the near future.
So even if we will not accept this PR, your help and the idea that has been grown out of it is much appreciated! Please continue to propose PRs and ideas as you desire. We much appreciate every input from the community!

PR #255, opened 2017-06-09
Created Gemfile to setup and run Jekyll using bundler

s1hofmann:
When I built the Citrus docs on my machine I had to go through an iteration of failing builds due to missing Jekyll plugins.
Using this Gemfile it is possible to automate this setup and run Jekyll locally using bundler.

In case bundler is not already installed:

    gem install bundler

Jekyll build / serve:

    bundle install
    bundle exec jekyll build
    bundle exec jekyll serve

### DIFF
--- a/src/site/Gemfile
+++ b/src/site/Gemfile
@@ -1,0 +1,9 @@
+# A Gemfile to build the Citrus Jekyll documentation
+source "https://rubygems.org"
+
+gem "jekyll", "~> 3.4"
+gem "jekyll-feed", "~> 0.9.2"
+gem "jemoji", "~> 0.8.0"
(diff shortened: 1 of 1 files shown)


christophd:
The site generation process is designed to run with jekyll docker image. you can call:

    mvn clean resources:resources package site-deploy -Psite,release-local -N

This will automatically start the jekyll docker container for site generation. as a result you have the generated site html sources in the local Maven target/site build folder. For local deployment call

    mvn docker:start -Psite -N

Wait for the Docker container to start and open the browser to http://localhost:4000
You will be able to see and navigate through the generated site.

s1hofmann:
With an even better solution in place, this PR is not really necessary. Closing.

PR #296, opened 2017-10-13
Updated HttpMessageContentBuilder to account for variable substitution

s1hofmann:
in headers.

The existing implementation did not account for variable substitution in header keys.
In Citrus < 2.7.2 variable substitution for maps only worked on entries, not keys, so after substituting values in headers, existing entries would be overwritten.
With variable substitution for keys in place this no longer works and previous headers have to be cleared, otherwise there'd be one header containing ${variables} and a separate one with actual values.

This in fact can cause HTTP 400 - Bad request errors.


svettwer:
Hi!

I've reviewed your changes and found out that header keys and values are already substituted. Nevertheless, I've added your test for regression test purposes.
Funny enough, this test showed, that to many headers are sent if a substitution in the header key was performed, because the substituted header including the variable expression has also been sent to the SUT.
I've also fixed this by replacing the header before sending instead of merging them.

Thx a lot for your contribution!

BR,
Sven

Issue #298, opened 2017-10-25
TestNG Dataprovider parameters cannot be used in <variables> block


gucce:
# Problem

When using a TestNG Dataprovider, the corresponding Citrus variables are only usable in the `<actions>` block but not in the `<variables>` block.

### TestNG Definition

```java
public class IT_Some_Test extends AbstractTestNGCitrusTest {

    @DataProvider(name = "someDataProvider")
    public Object[][] someDataProvider() {
        return new Object[][]{
                new Object[]{"some_value"},
        };
    }

    @Test(dataProvider = "someDataProvider")
    @CitrusParameters({"someVariable"})
    @CitrusXmlTest(name = "IT_Some_Test")
    @SuppressWarnings("unused")
    public void test(String someVariable) {
    }

}
```

### Citrus XML Test

```xml
<testcase name="IT_Some_Test">
    <variables>
        <!-- At this point the TestNG parameter 'someVariable' is not set yet. The test will fail. -->
        <variable name="otherVariable" value="${someVariable}"/>
    </variables>

    <actions>
        <create-variables>
            <!-- At this point the TestNG parameter 'someVariable' is set -->
            <variable name="otherVariable" value="${someVariable}"/>
        </create-variables>
    </actions>
</testcase>
```

# Solution

Set TestNG parameters in `TestContext` before any other variables are set (which may depend on TestNG parameters).

PR #315, opened 2017-12-06
298 test ng dataprovider parameters cannot be used in variables block

FMAOuroboros:
Hi!

I solved #298 by adding the TestNG DataProvider values as variables to the TestContext, as proposed in the Issue by @gucce.
In addition, I added the test case provided in the Issue as an automated test so that this problem will never occur again. I'm not sure whether I met the your expectations concerning the project structure, test coverage and code style.
I'm looking forward to your feedback.

BR,
Sven

### DIFF
--- a/modules/citrus-core/src/main/java/com/consol/citrus/context/TestContext.java
+++ b/modules/citrus-core/src/main/java/com/consol/citrus/context/TestContext.java
@@ -213,6 +213,26 @@ public class TestContext {
 
         variables.put(VariableUtils.cutOffVariablesPrefix(variableName), value);
     }
+
+    /**
+     * Add variables to context.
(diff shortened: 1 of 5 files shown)


christophd:
Thank you for the help! I reviewed the changes made and found out that the fix should be in testcase where the test parameters are now converted to variables before the actual test variables are created.

This enables us to use the parameters in variables as expected.

Also I moved the data provider test to the integration package because it loads the complete Citrus environment and is there fore of integration test nature compared to normal unit tests.

Thanks again for your help, it is very valuable.

FMAOuroboros:
Thx for your feedback!

I'll review your changes so that I get a better understanding of the project.

Issue #271, opened 2017-07-25
Variable in HTTP request within template not updated


eikeg:
A parameter of a template is shown correctly in an `echo` message, but not in an http request.

The following sample code creates **two times** the GET request with `id=123` (instead of `id=456`the second time):
```xml
	<template name="httpClient">
		<echo>
			<message>ID = ${id}</message>
		</echo>
		<http:send-request client="httpIccs">
			<http:GET path="/iccs-xsoap/version?id=${id}" />
		</http:send-request>
	</template>
	
			<call-template name="httpClient">
				<parameter name="id" value="123" />
			</call-template>
			<call-template name="httpClient">
				<parameter name="id" value="456" />
			</call-template>
```
This seems to apply to version 2.7.2 only.

PR #317, opened 2017-12-08
Update http request variable in template

FMAOuroboros:
Hi!

This is my attempt to solve #271. 
I found out, that the HttpMessageContentBuilder.buildMessageContent method worked on the initial request defined in the template and therefore replaced all variable placeholders with the concrete values. This led to the observed behavior. 

Now the buildMessageContent method works on a copy of the initial HttpMessage so that it is idempotent concerning the provided message. To create the copy, I had to adapt the copy Constructor of the HttpMessage, because the cookies where not copied if the provided Message was a HttpMessage instance.

In addition I applied some structural changes to the HttpMessageContentBuilder by extracting several methods. The non functional changes have been applied after the actual problem has been solved. So you may want to review commit by commit. 

As always: I'm looking forward to your feedback. ;-)

BR,
Sven

### DIFF
--- a/modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
+++ b/modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
@@ -50,6 +50,7 @@ public class HttpMessage extends DefaultMessage {
      */
     public HttpMessage(Message message) {
         super(message);
+        setCookies(message);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


christophd:
Thanks a lot! brilliant work here, had just some small cosmetic changes and made sure to use a copy of cookies, too.

FMAOuroboros:
Hi!

Thank you very much. I'm glad that you found my changes helpful.
It makes totally sense to create copies of the cookies as well. I think, I was a little bit too euphoric as the tests became green so that I did not thought enough about possible side effects. ;-)
I'll review your cosmetic changes to get a better understanding of the style you like in this project.

Issue #353, opened 2018-03-21
HttpMessageController is not thread-safe


imezhenkov:
Hi Christoph,

During the investigation of the instability of our functional tests I found that Citrus' http server sometimes returns wrong responses for concurrent requests. After digging deeper I revealed that the cause of problem lies in not thread-safe handing of response in HttpMessageController. There is a field `responseCache` which was introduced for message tracing reasons but which is also used to return response entity.
Possible solution here is to use local variable for return instead of the field and set `responseCache` from the variable.
Please find in the attachments the test to reproduces the issue.
Tested on version `2.7.3`.

[http_controller_test.zip](https://github.com/christophd/citrus/files/1832459/http_controller_test.zip)

![image](https://user-images.githubusercontent.com/37608332/37700470-50531dfc-2cf4-11e8-9645-8bfc104379f2.png)

christophd:
Fixed thanks to #354

PR #354, opened 2018-03-21
Update HttpMessageController.java to thread-safely handle response from endpoint adapter

imezhenkov:
The change fixes issue `HttpMessageController is not thread-safe #353`


christophd:
Thank you very much!

Issue #369, opened 2018-04-26
Default responses for JDBC requests


FMAOuroboros:
As a test designer, I want to specify default responses for jdbc requests, that are generated by the system under test but are irrelevant for the test itself, so that these irrelevant jdbc requests don't result in a failed tests.

PR #372, opened 2018-04-30
#369 Default responses for JDBC requests

joroboro:
- new configuration property 'auto-reply-connection-validation-queries' on jdbc-server introduced
- in case the flag 'auto-reply-connection-validation-queries' is set to true, certain queries are intercepted in the JdbcEndpointAdapterController and are automatically answered by Citrus with a success and an empty data set.
- a new citrus property 'citrus.jdbc.connection.validaiton.queries' introduced. It can be used to provide a list of queries, which has to be automatically answered by Citrus


christophd:
Thanks a lot for this change. I will think about the property name 'auto-reply-connection-validation-queries' though because I think it is too long. I am thinking of something like 'auto-handle-system-queries' or something.

joroboro:
I am also not really happy with the name, because of its length, but unfortunately I didn't found a better one. I think, that auto-handle-system-queries is too unspecific. May be we can find something, which is not so long but at the same time specific enough.

PR #488, opened 2018-09-26
Issue/418 thread leak

bwittwer:


svettwer:
Hi!

I reviewed your PR. It looks great!
At first I was a little sceptical concerning the necessity of the `FINISHER_THREAD_PREFIX` constant and the thread using that prefix, because the test you've implemented depends on the existence of this constant. Therefore the test is not executable before your changes. That being said, the test can not be taken as a proof of the former thread leak.
Nevertheless, I haven't found a way to proof the former thread leak without making too many assumptions concerning the actual thread setup in the JVM.

Long story short: Great work! Thanks a lot for your contribution! :+1: 

BR,
Sven

bwittwer:
Hi Sven,

You are right, I should have put some explanation in the PR comments about the testing.

I aggree that the test itself cannot be used as proof of the leak. But I also did not found a way that is simple (that can be executed at build time in a junit, independently of the jvm setup), deterministic (not statistical based on a larg number of execution) and automated (I used JProfiler to spot it) without instrumenting the code that way.

Happy to contribute :smiley: 

BR,
Bernard.

PR #500, opened 2018-10-05
Use 'contains' instead of 'equals' in order to process compound Conte…

Furiosina:
…nt-Type headers (e.g. application/x-www-form-urlencoded;utf-8)

### DIFF
--- a/modules/citrus-http/src/main/java/com/consol/citrus/http/servlet/CachingHttpServletRequestWrapper.java
+++ b/modules/citrus-http/src/main/java/com/consol/citrus/http/servlet/CachingHttpServletRequestWrapper.java
@@ -55,7 +55,7 @@ public class CachingHttpServletRequestWrapper extends HttpServletRequestWrapper 
 
         Map<String, String[]> params = new HashMap<>();
         if (RequestMethod.POST.name().equals(getMethod()) || RequestMethod.PUT.name().equals(getMethod())) {
-            if ("application/x-www-form-urlencoded".equals(getContentType())) {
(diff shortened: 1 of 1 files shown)


modules/citrus-http/src/main/java/com/consol/citrus/http/servlet/CachingHttpServletRequestWrapper.java
```diff
@@ -55,7 +55,7 @@ public CachingHttpServletRequestWrapper(HttpServletRequest request) {
 
         Map<String, String[]> params = new HashMap<>();
         if (RequestMethod.POST.name().equals(getMethod()) || RequestMethod.PUT.name().equals(getMethod())) {
-            if ("application/x-www-form-urlencoded".equals(getContentType())) {
+            if (getContentType() != null && getContentType().contains("application/x-www-form-urlencoded")) {
```
svettwer: `.contains()` is unfortunately insufficient for the check.
This would also allow `foo-application/x-www-form-urlencoded-bar`.
One could think about using `getContentType().startsWith("application/x-www-form-urlencoded")` here.

Furiosina: Thank you for your response! 
Now I see that it is not that simple. I have added an issue. Hope it would be fixed. 
I'm afraid I don't have enough competence to correct you about RFC2045 specification, but my "system under test" (which is springboot based application) sends exactly such Content-Type header in requests.

Issue #513, opened 2018-10-15
Citrus Build fails with Java 11


sparsick:
**Citrus Version**
2.8.0-Snapshot
OpenJDK 11
Maven 3.5.2

**Expected behavior**
Build is successful

**Actual behavior**
First fail because of javadoc-maven-plugin issue:
````
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.10.4:jar (attach-javadocs) on project citrus: Execution attach-javadocs of goal org.apache.maven.plugins:maven-javadoc-plugin:2.10.4:jar failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-javadoc-plugin:2.10.4:jar: java.lang.ExceptionInInitializerError: null
````

This can be fixed with a version upgrade of javadoc-maven-plugin to 3.0.1.

Second fail because of jaxb2-maven-plugin:
````
[ERROR] Failed to execute goal org.codehaus.mojo:jaxb2-maven-plugin:2.4:xjc (default) on project citrus-model-core: Execution default of goal org.codehaus.mojo:jaxb2-maven-plugin:2.4:xjc failed: A required class was missing while executing org.codehaus.mojo:jaxb2-maven-plugin:2.4:xjc: javax/activation/DataSource
````

I modified the configuration of the jaxb2-maven-plugin, so that it is use javax-activation dependency:
````
        <plugin>
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>jaxb2-maven-plugin</artifactId>
          <version>2.4</version>
          <dependencies>
            <dependency>
              <groupId>javax.activation</groupId>
              <artifactId>activation</artifactId>
              <version>1.1.1</version>
            </dependency>
          </dependencies>
        </plugin>
````

But then I get the third fail because of SaxParseException:

````
[INFO] --- jaxb2-maven-plugin:2.4:xjc (default) @ citrus-model-core ---
[INFO] Created EpisodePath [/home/sparsick/dev/workspace/citrus/modules/citrus-model/model-core/target/generated-sources/jaxb/META-INF/JAXB]: true
[INFO] Created EpisodePath [/home/sparsick/dev/workspace/citrus/modules/citrus-model/model-core/target/generated-sources/jaxb/META-INF/JAXB]: true
WARNING: An illegal reflective access operation has occurred
[...]

PR #515, opened 2018-10-16
Build improvement - add more jdk

sparsick:
Enable multiple job in travisci so that problem like #513 can be discovered earlier

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk9
+  - oraclejdk10
+  - openjdk10
+  - openjdk11
 
 dist: trusty
 sudo: required


svettwer:
Hi!

When I understood #513 correctly, #513 is about using Citrus in a Java 11 environment, not building Citrus with java 9 or higher. Building with java 9 or higher is currently not supported, because we haven't started with the migration yet.

Using Citrus in Java 11 would be a different use case that building it with java 11.

BR,
Sven

sparsick:
My point in #513 is that I cannot build the citrus project with java 11.

svettwer:
Hi!

Thx for clarification. Building Citrus with java 9 or higher is in scope but currently not scheduled. The adaptations to the pipeline will be taken into account as soon as the migration starts.

BR,
Sven

Issue #531, opened 2018-11-14
Travis+Sonarcloud -> External PR error


svettwer:
Hi!

After adding an automatic sonarcloud analysis to the project, builds of external pull requests will not pass as #527 shows.

The reason for this is the limited security context of travis for builds from external branches.
This issue is known and a workaround is proposed but untested.

- [Forum report](https://community.sonarsource.com/t/travis-plugin-is-failing-on-external-pull-request/807/5)
- [Jira issue](https://jira.sonarsource.com/browse/TRAVIS-19)

```bash
if [ "$TRAVIS_SECURE_ENV_VARS" = true ]; then
   mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar
else
  mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent install
fi
```

BR,
Sven

svettwer:
The build log shows that the script was not functional in the travis environment.

I've found another section in the travis documentation about [pull requests and security restricitons](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions).
This section contains the following snippet:

```bash
if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./travis/run_on_pull_requests; fi
if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then bash ./travis/run_on_non_pull_requests; fi
```

I'll alter the introduced script based on this new information.

svettwer:
The condition I've posted is working but also prevents internal PRs.
The last commit contains a promising condition.

svettwer:
Merged and works as expected for external PRs [(log)](https://travis-ci.com/citrusframework/citrus/builds/91548774#L454)

PR #534, opened 2018-11-15
Citrus 531

svettwer:
Hi!

This PR is concerning #531. I've implemented a switch to distinguish internal an external builds. This is currently required because of a bug in the sonarcloud travis plugin. 
I haven't had the chance to test it with an external PR but it's working for internal builds as this [log](https://travis-ci.com/citrusframework/citrus/jobs/158742969#L465) shows.

I'd rebase an external pull request on this change as soon as it is in master to check if it works as expected.

BR,
Sven


svettwer:
Okay, the condition also excludes internal PR's

I'll have a look

PR #543, opened 2018-11-25
Fix overwritten cookie in HttpMessageConverter

kezhuw:
Uses `HttpHeaders.add` to append cookie name-value pair instead of `HttpHeaders.set` which dropping old `Cookie` header.


svettwer:
Hi!

Thx for your suggestion. :+1: 
Did you experience any problem with the current implementation? If so, it would be great if you could open an issue for that so that we've a little bit more context about the change, even if it's a one liner. :wink: 

BR,
Sven

kezhuw:
@svettwer See #561 (Lost cookie entries except the last one).

svettwer:
Hi!

I've created a internal branch (https://github.com/citrusframework/citrus/tree/kezhuw-fix-overwritten-cookies) and added a integration test to prevent regression. 
Your change fixed the issue :+1:. Thx a lot for your contribution.

I'll now add some unit tests on the `HttpMessageConverter` and create a internal PR, so that my changes are reviewed as well.

This PR will be automatically closed as soon as the changes are merged into master.

BR,
Sven

Issue #561, opened 2018-12-01
Lost cookie entries except the last one


kezhuw:
**Citrus Version**
2.7.5

**Expected behavior**
Server gets all cookie entries.

**Actual behavior**
Server gets only last cookie entry.

**Test case sample**

```java
JUnit4CitrusTestDesigner d = new JUnit4CitrusTestDesigner();
d.http().client("clientId")
    .send()
    .delete("/path")
    .cookie(new Cookie("a", "a"))
    .cookie(new Cookie("b", "b"))
```

#543 try to solve this.

svettwer:
Hi!

It seems that this issue is at lest linked, if not a duplicate of #455. It could be the case that the root cause is the same but with different symptoms in both issues.

I've extended the test case so that it's executable as a integration test.
```java
public class TodoListIT extends TestNGCitrusTestRunner {

    @Autowired
    private HttpClient httpClient;

    @Autowired
    private HttpServer httpServer;

    @Test
    @CitrusTest
    public void testGet() {

        final Cookie aCookie = new Cookie("a", "a");
        final Cookie bCookie = new Cookie("b", "b");

        http(http -> http.client(httpClient)
                .send()
                .delete("/path")
                .cookie(aCookie)
                .cookie(bCookie));

        http(http -> http.server(httpServer)
                .receive()
                .delete("/path")
                .cookie(aCookie)
                .cookie(bCookie));
    }
}
```

It fails with
```
com.consol.citrus.exceptions.TestCaseFailedException: Validation failed: No matching header for key 'citrus_http_cookie_a'
	at com.consol.citrus.TestCase.executeAction(TestCase.java:227)
	at com.consol.citrus.dsl.runner.DefaultTestRunner.run(DefaultTestRunner.java:188)
	at com.consol.citrus.dsl.runner.DefaultTestRunner.http(DefaultTestRunner.java:540)
	at com.consol.citrus.dsl.testng.TestNGCitrusTestRunner.http(TestNGCitrusTestRunner.java:375)
	at com.consol.citrus.samples.todolist.TodoListIT.testGet(TodoListIT.java:53)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #563, opened 2018-12-03
561: Lost cookie entries except the last one + Tests and refactoring

svettwer:
Hi!

This PR is based on PR #543 and issue #561 of @kezhuw.

I've added the integration test from the issue to proof, that the fix from the original PR solves the issue of overwritten cookies. In addition, I've added unit tests to the `HttpMessageConverter` based on the current implementation. The only thing I've changed implementation wise is the handling of the `Secure` flag for Cookies, based on the [Mozilla Set-Cookie documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie).

The old implementation expected the Secure flag to be `Secure=true`. I know that the fix is not ideal but I didn't want to change the contract of `getCookieParam`. :sweat_smile: 

To the reviewer: If you want to merge this directly after review, please merge the changes into the `v2.7-bugfix` branch as well or just approve and I'll do the merging. Thx! :+1: 

BR,
Sven

### DIFF
--- a/modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
+++ b/modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
@@ -18,6 +18,7 @@ package com.consol.citrus.http.message;
 
 import com.consol.citrus.exceptions.CitrusRuntimeException;
 import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
 
 import javax.servlet.http.Cookie;
 import java.util.LinkedList;
@@ -42,7 +43,7 @@ class CookieParser {
     Cookie[] convertCookies(HttpEntity<?> httpEntity) {
         final List<Cookie> cookies = new LinkedList<>();
 
-        List<String> inboundCookies = httpEntity.getHeaders().get("Set-Cookie");
+        List<String> inboundCookies = httpEntity.getHeaders().get(HttpHeaders.SET_COOKIE);
         if (inboundCookies != null) {
             for (String cookieString : inboundCookies) {
                 Cookie cookie = convertCookieString(cookieString);
@@ -108,7 +109,7 @@ class CookieParser {
         }
 
         if(SECURE.equals(param) && cookieString.contains(SECURE)) {
-            return "true";
+            return String.valueOf(true);
         }
 
         if (cookieString.contains(param + '=')) {
--- a/modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
+++ b/modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
@@ -1,0 +1,127 @@
+/*
+ *    Copyright 2018 the original author or authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.consol.citrus.http.message;
+
+import com.consol.citrus.exceptions.CitrusRuntimeException;
+import org.springframework.http.HttpEntity;
+
+import javax.servlet.http.Cookie;
+import java.util.LinkedList;
+import java.util.List;
+
+class CookieParser {
+
+    private static final String NAME = "Name";
+    private static final String VALUE = "Value";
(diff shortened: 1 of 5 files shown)


modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
```diff
@@ -0,0 +1,127 @@
...
+        if (param.equals(VALUE)) {
+            if (cookieString.contains(";")) {
+                return cookieString.substring(cookieString.indexOf('=') + 1, cookieString.indexOf(';'));
+            } else {
+                return cookieString.substring(cookieString.indexOf('=') + 1);
+            }
+        }
+
+        if(SECURE.equals(param) && cookieString.contains(SECURE)) {
+            return "true";
```
s1hofmann: A bit verbose, but could be replaced with `String.valueOf(true)`

modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessageConverter.java
```diff
@@ -17,158 +17,99 @@
...
+        HttpHeaders httpHeaders = createHttpHeaders(httpMessage, endpointConfiguration);
 
         Object payload = httpMessage.getPayload();
         if (httpMessage.getStatusCode() != null) {
             return new ResponseEntity<>(payload, httpHeaders, httpMessage.getStatusCode());
         } else {
             for (Cookie cookie : httpMessage.getCookies()) {
-                httpHeaders.set("Cookie", cookie.getName() + "=" + context.replaceDynamicContentInString(cookie.getValue()));
+                httpHeaders.add(
+                        "Cookie",
```
s1hofmann: We could use `HttpHeaders.COOKIE` constant instead of a String.

modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessageConverter.java
```diff
@@ -17,158 +17,99 @@
...
-        HttpMessage httpMessage = new HttpMessage(message.getBody() != null ? message.getBody() : "", convertHeaderTypes(mappedHeaders));
+        HttpMessage httpMessage = new HttpMessage(extractMessageBody(message), convertHeaderTypes(mappedHeaders));
 
         for (Map.Entry<String, String> customHeader : getCustomHeaders(message.getHeaders(), mappedHeaders).entrySet()) {
             httpMessage.setHeader(customHeader.getKey(), customHeader.getValue());
         }
 
         if (message instanceof ResponseEntity<?>) {
             httpMessage.status(((ResponseEntity<?>) message).getStatusCode());
-            httpMessage.version("HTTP/1.1"); //TODO check if we have access to version information
```
s1hofmann: So there's no version info? 😃

modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
```diff
@@ -0,0 +1,127 @@
...
+        return cookie;
+    }
+
+    /**
+     * Extract cookie param from cookie string as it was provided by "Set-Cookie" header.
+     * @param param The parameter to extract from the cookie string
+     * @param cookieString The cookie string from the cookie header to extract the parameter from
+     * @return The value of the requested parameter
+     */
+    private String getCookieParam(String param, String cookieString) {
```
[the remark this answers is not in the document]
svettwer: Thought about this as well but somehow I thought this is only available since 1.9.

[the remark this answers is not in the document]
svettwer: Ah! Now i got it.

[java.net.HttpCookie#parse](https://docs.oracle.com/javase/8/docs/api/java/net/HttpCookie.html#parse-java.lang.String-)  is a different type. So I'd have to migrate the cookie type through the application. I'd suggest to open a separate issue for a replacement as a maintenence task. Would you be okay with this?

modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
svettwer: Yeah, well... :thinking: 
I'll do it, but I think the intended use case of `String.valueOf()` is more like if you have a variable parameter so that's not clear whether the boolean value will be `true` or `false` (e.g. a recursive set) but with String as return type. :smiley:

modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieParser.java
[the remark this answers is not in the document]
svettwer: I double checked this option this morning. There is a lot of code to be touched and a lot of tests to be written for this. I'd definitely go with a low maintenance ticket.

modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessageConverter.java
svettwer: I've added some comments to the code to make this clear. :wink:

PR #567, opened 2018-12-12
Modify ReceiveMessageBuilder to accept Maps of assertions for validation

tvelkoff:
I propose to add two new methods to ReceiveMessageBuilder in the Java DSL: validateXpath and validateJsonPath, which accept Maps as arguments. (see lines 585-604)

Here is the motivation... I have many test cases similar to this:

```java
designer.soap()
  .client(this.client)
  .receive()
  .schemaValidation(false)
  .validate("//lexs:ResultCode", "Success")
  .validate("//lexs:SearchResponseMetadata/lexs:MaxItemMatchesRequested", "10")
  .validate("//lexs:SearchResponseMetadata/lexs:NumberItemMatches", "10")
  .validate("//lexs:SearchResponseMetadata/lexs:ServerLimitIndicator", "true");
```
where the only difference is a variable number of assertions. What motivates the need for the new methods is the desire to create a single method in my test code that eliminates redundancy in individual tests, like this:

```java
protected void runTest(TestDesigner designer, String inputPathString, Map<String, Object> assertions) {
...
  designer.soap()
  ... (as above)
  .validate(assertions); <--new method used here
}
```

I found no test cases for this class and therefore created a JUnit 5 test class that covers the public methods as well as the two proposed new methods. There are four tests I left in Disabled state - I was unable to get the desired outcome using Mockito.

There are two other code changes that are unrelated, and that could be rejected or deferred.
`getScriptValidationContext` throws a `NullPointerException` if `messageType` has not previously been called. For this I added an assertion similar to those on some other methods.

A couple of methods related to `headerFragment` proved difficult to mock and test because a `StringResult` is created inside the method. I introduced helper methods that could be called by the unit tests, but left the existing public method signatures unchanged. These methods delegate to the helpers. Again, these changes were introduced only to facilitate unit testing.

I have just recently found Citrus - thank you for all of this work. I have been working in this problem space for about ten years and have needed a tool like this. Vielen Dank!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
     <activemq.version>5.15.4</activemq.version>
     <vertx.version>3.5.4</vertx.version>
     <groovy.version>2.5.0</groovy.version>
-    <jetty.version>9.4.11.v20180605</jetty.version>
+    <jetty.version>9.4.15.v20190215</jetty.version>
     <testng.version>6.11</testng.version>
     <junit.version>4.12</junit.version>
     <junit.jupiter.version>5.3.2</junit.jupiter.version>
--- a/src/manual/changes-2-7.adoc
+++ b/src/manual/changes-2-7.adoc
@@ -19,6 +19,7 @@ https://github.com/citrusframework/citrus/issues/583[#583 - Param list support w
 [[changes-security-2-7-10]]
 === Security fixes
 * https://github.com/citrusframework/citrus/issues/612[#612 - Update org.springframework.ws.spring-xml to 3.0.6 or later]
+* https://github.com/citrusframework/citrus/issues/639[#639 - Update jetty-server to 9.4.12.v20180830 or later]
 
 [[changes-2-7-9]]
 == Changes in Citrus 2.7.9


svettwer:
Hi @tvelkoff!

Thx for your PR. :+1: 
I'll have a look as soon as possible and provide some feedback.

It would be great if you could open an issue concerning your PR in the meantime, so that we can separate the discussion about the business logic (issue) from the technical discussion (PR). I'd also have to check why the Travis build did not start.

BR,
Sven

svettwer:
Hi!

I checked why the build is not passing. We've a missing dependency here for the `MockitoExtension`. In addition, an update of mockito to `>= 2.20` was required, so that `lenient` becomes available.

I created a branch within the project called `enhancement/571/validateWithMapParameter` to apply some changes.

BR,
Sven

svettwer:
Wow this is a real bunch of unit tests! Thank you so much! :+1: :tada:

tvelkoff:
Sorry, I made an error trying to push something to my own GitHub. I used the wrong remote label.
I hope I didn't accidentally close this as well.

svettwer:
Hi!

I was indeed curious what led to the force push and the closed PR. ;-)

Everything is fine from my perspective. Your changes are part of my internal branch, so nothing is lost. Nevertheless, if I would not have my internal review branch, your changes would be gone. `git push --force` is a dangerous tool. =)

BR, 
Sven

tvelkoff:
No kidding! Really sorry about that.
I was attempting to force my own copy of Citrus to synch up with 2.9.0-SNAPSHOT.
Yikes.

svettwer:
No worries! Everything is fine. I merged the previous content of this PR into master today. =)
Thanks a lot for your help and the awesome amount of tests you contributed! :+1:

BR, 
Sven

tvelkoff:
I wish I could have eliminated all of the yellow warnings in Eclipse related to generic parameters. Despite reading Angelika Langer's blogs about generics, there are still puzzlers I can't solve. :(

svettwer:
Some of the generic issues are not that easy to solve and sometimes require some changes of the types in general. That is nothing you can do just like that. Nevertheless, you applied the scouts rule and left the place you visited cleaner as it was before. That is great and very helpful! =)

Issue #569, opened 2018-12-13
duplicated cookie entries in http request


svettwer:
Extracted from #544 
------------------

**Citrus Version**
2.7.8

**Actual behavior**
When I setup a HTTP message using the HttpMessageContentBuilder, added cookies are doubled during processing.

**Expected behavior**
When I setup a HTTP message using the HttpMessageContentBuilder, added cookies are unique in the message after processing.

**Test case sample**
```java
    @Test
    public void testTemplateHeadersArePreserved(){

        //GIVEN
        final HttpMessage msg = new HttpMessage("testPayload");
        msg.setCookies(new Cookie[]{new Cookie("foo","bar")});

        final HttpMessageContentBuilder builder =
                new HttpMessageContentBuilder(msg, new StaticMessageContentBuilder(msg));

        //WHEN
        final HttpMessage builtMessage = (HttpMessage) builder.buildMessageContent(
                new TestContext(),
                String.valueOf(MessageType.XML));

        //THEN
        Assert.assertEquals(builtMessage.getCookies().size(), 1);
        Assert.assertEquals(builtMessage.getCookies().get(0).getName(), "foo");
        Assert.assertEquals(builtMessage.getCookies().get(0).getValue(), "bar");
    }
```

This test would fail as the size of `builtMessage.getCookies()` would be 2 due to the doubling. The root cause is located in the `HttpMessage` implementation of `setCookies`.
#544 contains a fix for this.

BR,
Sven

svettwer:
Todo: Add some tests for `HttpMessageContentBuilder`

PR #544, opened 2018-11-25
Fix duplicated cookie entries in http request

kezhuw:
Second overloading method `HttpMessageContentBuilder.buildMessageContent` uses
`HttpMessage.setCookies` to set cookies for message derived from
template without clearing old cookies from template.

This merge request fixes this by clearing old cookies in
`HttpMessage.setCookies`.

Feel free to reject this merge request if semantics change of `HttpMessage.setCookies` is harmful. Or should I degrade this merge request to an issue report ?

PS: I found only one usage of `HttpMessage.setCookies` in `HttpMessageController.handleRequestInternal` and didn't saw the harmness.

PR #572, opened 2018-12-17
fix/569/duplicated cookie entries in http request

svettwer:
This fix is concerning #569 which is based on #544.

I've added some tests and documentation to the original changes.

### DIFF
--- a/modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieConverter.java
+++ b/modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieConverter.java
@@ -19,12 +19,21 @@ package com.consol.citrus.http.message;
 import com.consol.citrus.exceptions.CitrusRuntimeException;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
+import org.springframework.util.StringUtils;
 
 import javax.servlet.http.Cookie;
+import java.io.Serializable;
 import java.util.LinkedList;
 import java.util.List;
 
-class CookieParser {
+/**
+ * Class to convert Objects from or to Cookies
+ *
+ * This class should be replaced as soon as possible by a third party cookie parser
+ * The implementation of the Serializable interface is cause by the {@link HttpMessage} implementation of Serializable
+ * and the implications from that.
+ */
+class CookieConverter implements Serializable {
 
     private static final String NAME = "Name";
     private static final String VALUE = "Value";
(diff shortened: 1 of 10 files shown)


modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieConverter.java
```diff
@@ -54,6 +63,45 @@
...
+        }
+
+        if (StringUtils.hasText(cookie.getComment())) {
+            builder.append(";" + COMMENT + "=").append(cookie.getComment());
+        }
+
+        if (cookie.getSecure()) {
+            builder.append(";" + SECURE);
+        }
+
```
martinmaher: What about HttpOnly flag?

modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
```diff
@@ -416,55 +438,23 @@ public void setCookies(Cookie[] cookies) {
 
     /**
      * Adds new cookie to this http message.
-     * @param cookie
-     * @return
+     * @param cookie The Cookie to set
+     * @return The altered HttpMessage
      */
     public HttpMessage cookie(Cookie cookie) {
         this.cookies.add(cookie);
```
martinmaher: Does it make sense here to overwrite cookies with same name (i.e. check using name if already in list and overwrite rather than allowing multiple cookies with the same name)?

modules/citrus-http/src/test/java/com/consol/citrus/http/message/CookieConverterTest.java
```diff
@@ -150,31 +146,111 @@ public void testCookiePathIsPreserved(){
...
-        final Cookie[] cookies = cookieParser.convertCookies(responseEntity);
+        final Cookie[] cookies = cookieConverter.convertCookies(responseEntity);
 
         //THEN
-        assertEquals(1, cookies.length);
         assertTrue(cookies[0].getSecure());
     }
 
     @Test
     public void testCookieVersionIsPreserved(){
```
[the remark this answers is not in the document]
svettwer: :+1:

modules/citrus-http/src/main/java/com/consol/citrus/http/message/CookieConverter.java
svettwer: Good point! `HttpOnly` was missing the whole time.

modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
svettwer: Well, this makes sense from my perspective as well but `cookie(Cookie cookie)` is a method exposed to the public API via a builder with the documentation of *"Adds new cookie to this http message."*.

So I'd not change this logic here. In addition, there is no technical or functional requirement for that change. Maybe this is something to consider for future major releases.

martinmaher: AFAIK there is a requirement from a technical point of view (https://tools.ietf.org/html/rfc6265#section-4.1.1) 

> Servers SHOULD NOT include more than one Set-Cookie header field in
> the same response with the same cookie-name.  (See Section 5.2 for
> how user agents handle this case.)

svettwer: Okay, you've convinced me. :smile:

Issue #535, opened 2018-11-16
Citrus functions - Inconsistent values of global variables


svettwer:
Extracted from [stackoverflow](https://stackoverflow.com/questions/53222826/global-random-number-for-all-test-cases)
-------------

**Citrus Version**
Traced it back to 2.0

**Expected behavior**
Accessing a global variable containing a Citrus function such as  `<variable name="uuid" value="citrus:randomUUID()"/>` twice, returns the same value.

**Actual behavior**
Accessing a global variable twice returns different values.

**Test case sample**
A `GlobalVariables` java config
```java
@Configuration
public class EndpointConfig {

    @Bean
    public GlobalVariables globalVariables(){
        final GlobalVariables globalVariables = new GlobalVariables();
        globalVariables.getVariables().put("myVariableName", "citrus:randomUUID()");
        return globalVariables;
    }
}
```

**OR**

A `citrus-context.xml` config
```xml
    <citrus:global-variables>
        <citrus:variable name="myVariableName" value="citrus:randomUUID()"/>
    </citrus:global-variables>
```

with the test definition

```java
    @Test
    @CitrusTest
    public void testGet() {
        echo("${myVariableName}");
        echo("${myVariableName}");
    }
```
**OR**

```xml
    <actions>
      <echo>
        <message>${myVariableName}</message>
      </echo>
      <echo>
        <message>${myVariableName}</message>
      </echo>
    </actions>
```
results in

```
0:49:57,904 INFO         citrus.Citrus| TEST STEP 1/2: echo
10:49:57,906 INFO    actions.EchoAction| a110191a-0cfb-4cdc-a5a0-f018a8a51219
10:49:57,908 INFO         citrus.Citrus| TEST STEP 2/2: echo
10:49:57,909 INFO    actions.EchoAction| a043dd62-3def-4d27-bcd4-638e9cfae125
```

**Workaround to generate Random numbers**
```java
public class InitializationTest extends TestDesignerBeforeSuiteSupport {

    @Override
    public void beforeSuite(TestDesigner designer) {
        designer.createVariable("myGlobalNumber", RandomNumberFunction.getRandomNumber(3, true));

        designer.action(new AbstractTestAction() {
            @Override public void doExecute(TestContext testContext) {
                testContext.getGlobalVariables()
                        .put("myGlobalNumber", testContext.getVariable("myGlobalNumber"));
            }
        });

    }
}
```

PR #573, opened 2018-12-20
(#535) dynamic content (functions or references to other variables) i…

martinmaher:
…s resolved in global variables when added to the test context. This ensures that the value of the global variable remains constant throughout the execution of the test. In particular it fixes the problem where a citrus:randomNumber function was used as the value of a global variable; each time the global variable was referenced or read then a new value was returned.


modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        assertVariableValuesMatchWhenReadMultipleTimes(context, "globalSum1");
+
+        runner.echo("globalSum2: ${globalSum2}");
+        assertVariableValuesMatchWhenReadMultipleTimes(context, "globalSum2");
+
+        assertVariableValuesMatch(context, "globalSum1", "globalSum2");
+    }
+
+    private void assertVariableValuesMatchWhenReadMultipleTimes(final TestContext testContext, final String name) {
+        String val1 = testContext.getVariable(name);
```
svettwer: I think this method does not test what you want it to test.
`testContext.getVariable(name);` gets you the unresolved value of the variable. No dynamic replacement of functions or variables has been applied here.

Even if I rollback your changes and run the tests, they pass.
Here you can see why:
image
`val1` and `val2` both contain `citrus:randomNumber(5)` what makes them equal.

```suggestion
        String val1 = testContext.resolveDynamicValue(testContext.getVariable(name));
```
If you use this also for `val2`, the test fails as expected before you applied your change.
image

When I now add back your changes to the code, the tests pass as they should.

modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        String val2 = testContext.getVariable(name);
+        if (val1.equals(val2)) {
+            log.debug(String.format("Values match for variable %s. Value: %s", name, val1));
+        } else {
+            throw new RuntimeException(String.format("Values don't match for variable %s. Value1: %s, Value2: %s", name, val1, val2));
+        }
+    }
+
+    private void assertVariableValuesMatch(final TestContext testContext, final String name1, final String name2) {
+        String val1 = testContext.getVariable(name1);
```
svettwer: Same issue as for `assertVariableValuesMatchWhenReadMultipleTimes`

```suggestion
        String val1 = testContext.resolveDynamicValue(testContext.getVariable(name1));
```

modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        String val2 = testContext.getVariable(name2);
```
svettwer: Same issue as for `assertVariableValuesMatchWhenReadMultipleTimes`

```suggestion
        String val2 = testContext.resolveDynamicValue(testContext.getVariable(name2));
```

modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        runner.echo("globalRandomNumber1: ${globalRandomNumber1}");
+        assertVariableValuesMatchWhenReadMultipleTimes(context, "globalRandomNumber1");
+    }
+
+    @CitrusTest
+    @Parameters({"runner", "context"})
+    public void multipleGlobalVariablesWithCrossReferencesAndFunctions_readMultipleTimes_valueShouldRemainConstantOnEachRead(
+            @Optional @CitrusResource TestRunner runner, @Optional @CitrusResource TestContext context) {
+        runner.echo("globalSum1: ${globalSum1}");
+        assertVariableValuesMatchWhenReadMultipleTimes(context, "globalSum1");
```
svettwer: In my opinion, this is redundant to line [60](https://github.com/citrusframework/citrus/pull/573/files#diff-632f5ba22ee89f7d193923dd4e328331R60), because `globalSum1` is just a placeholder for a Citrus function just as `globalRandomNumber1` is. The same goes for `globalSum2`. So I think it would be enough to test for matching variables values here.

modules/citrus-core/src/main/java/com/consol/citrus/types/Tuple.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.consol.citrus.types;
+
+/**
+ * Very basic tuple implementation. No support for equals or hashcode so be careful when using this in collections
+ */
+public class Tuple<A, B> {
```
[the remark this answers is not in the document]
martinmaher: Agreed. I was just wary of introducing yet another dependency to citrus. I checked javatuples and right now it has no transitive dependencies ;-)

modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
martinmaher: Well spotted. I assumed that when using the citrus java dsl a call to getVariable would return the value with any dynamic content resolved. I incorporated your suggested change. 
On a side note - what is the recommended way to access variables using the java DSL in citrus? Calling resolveDynamicValue is not intuitive, at least it wasn't for me ;-)

modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
martinmaher: Agreed. I can't remember why i thought this necessary at the time.

modules/citrus-core/src/main/java/com/consol/citrus/context/TestContext.java
```diff
@@ -260,19 +298,40 @@ public void addVariables(Map<String, Object> variablesToSet) {
...
+        final K adaptedKey;
+
+        if (key instanceof String) {
+            adaptedKey = (K) replaceDynamicContentInString((String) key);
+        } else {
+            adaptedKey = key;
+        }
+
+        final V adaptedValue;
+        if (value instanceof String) {
```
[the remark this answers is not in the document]
martinmaher: NP, I'll change this.

modules/citrus-integration/src/it/java/com/consol/citrus/javadsl/runner/GlobalVariablesTestRunnerIT.java
svettwer: Well, now that you ask, I think that this is the only way to do it currently. :thinking: 
As a review of the current `TestContext` concept is planned for the next major, this should be addressed then.

svettwer:
PR concerning #535

Issue #553, opened 2018-11-30
Scp client builder does not offer autoReadFiles anymore


svettwer:
**Citrus Version**
    >= 2.7.7

**Description**
If you upgrade your Citrus version to 2.7.7 or higher, we've a breaking change in the `ScpClientBuilder` which does not support the `autoReadFiles` method anymore. We'll correct this with one of the future releases to ensure effortless version upgrades

**API before change**
```java
    /**
     * Sets the auto read files property.
     * @param autoReadFiles
     * @return
     */
    public ScpClientBuilder autoReadFiles(boolean autoReadFiles) {
        endpoint.getEndpointConfiguration().setAutoReadFiles(autoReadFiles);
        return this;
    }
```

**API after change**
Method not present anymore

**Additional information**
Issue: #435
Commit: https://github.com/citrusframework/citrus/commit/b64e2e5bf510809d622695a0c399836fe3ee7aab#diff-7b734d11520ab62bc8fe2b60dc0942d4L72

BR,
Sven

PR #593, opened 2019-01-16
(#553) Added ScpClientBuilderTest

svettwer:


svettwer:
I've discussed the changes with martin via phone. He approved, so I'll merge without explicit approve via github.

Issue #549, opened 2018-11-29
Fix public API breaking changes in 2.7.x


svettwer:
We've found some breaking changes in the 2.7.x release family.

This issue contains a list of breaking changes that will be fixed with 2.7.9 to ensure a effortless migration to this version.

All reconstructed methods will be annotated as `@Deprecated` and removed with an upcoming 3.0.0 release.

- [x] #547 - Breaking change in waitFor().http() public API
- [x] #551 - Breaking change in selenium().browser().type() public API
- [x] #552 - MessageContentBuilder interface broken for buildMessageContent
- [x] #553 - Scp client builder does not offer autoReadFiles anymore
- [x] #554 - Change of class signature and typing of WaitConditionBuilder
- [x] #555 - WaitBuilder.condition() changed return type
- [x] #556 - Breaking change in waitFor().message()
- [x] #557 - Breaking change in waitFor().file(String)
- [x] #558 - Breaking change in waitFor().file(File)

christophd:
Honestly, most of the changes listed here are not visible to the end user that just use the Citrus XML or Java DSL. Unless you have custom extensions in your project that deeply customize the behavior of the framework you won't even notice some of these changes. 

What is the motivation behind this? Are there any end users complaining about too many breaking changes in public APIs in previous versions? I can not remember to have received such complains in the past.

I think introducing a deprecated duplication of what has already been changed in order to remove it in the next version (probably soonish v.2.8.0) is not worth the effort.

christophd:
How long are you going to keep the introduced deprecations? Keeping the deprecation for #511 for instance up to v3.0.0 is way too long in my opinion. I suppose to remove most or all of the introduced deprecations with v2.8.0.

svettwer:
Hi! 

>What is the motivation behind this? Are there any end users complaining about too many breaking changes in public APIs in previous versions?

Yes, we recently had complaints by users about such changes.

>I think introducing a deprecated duplication of what has already been changed in order to remove it in the next version (probably soonish v.2.8.0) is not worth the effort.
[...]
How long are you going to keep the introduced deprecations? Keeping the deprecation for #511 for instance up to v3.0.0 is way too long in my opinion.

We're going to remove the deprecated parts of the API with 3.0.0 as we're now following a [semantic versioning](https://semver.org/). In semantic versioning, incompatible API changes are only allowed when upgrading a major version.

BR,
Sven

christophd:
[...]

PR #594, opened 2019-01-16
Add WaitBuilderTest

svettwer:
Hi!

This PR intends to fix the broken quality gate.
During the reconstruction of the `WaitBuilder` API in #549, the quality gate broke because the reconstruction was now handles as untested *new code*.
Therefore I added test coverage to this class.

I also deprecated the `ms` methods which were a duplicated of `milliseconds`.


svettwer:
Thx for the response! :+1: 

Yeah, from the perspective that they add logic, I agree with you but  the `equals` and `hashCode`methods are generated by the IDE and not implemented by myself. These methods are just the regular java boiler plate. Testing those does not add any value at all, in my opinion.

BR,
Sven

PR #622, opened 2019-02-17
Improve Maven integration test execution

christophd:
Renamed context injection tests to follow only integration test naming pattern. Before that the tests were executed in both test and integration test phase through both surefire and failsafe Maven plugin.

Also put failure tests that fail on purpose to separate failsafe execution in Maven (via test groups). With this in place build error analysis gets more straight forward as the tests failing on purpose are not logging the failure state to the Citrus test summary. You will then just see the build error causing test failure in the summary.


christophd:
The failed code analysis is related to the Citrus tests not having any assertions. This basically applies to all Citrus tests.

Only thing I can do about it is to manually suppress the warning. 

Maybe disable that rule?

svettwer:
Hi!

Sorry, haven't had the chance to take a look. I would go with suppressing these warnings. Disabling the rule in general is not an option from my perspective.

BR,
Sven

svettwer:
Is it possible for you to suppress the warnings? :thinking: 
I think it should be if you've a sonarcloud account right?

christophd:
I can do this on a class level in the code base

christophd:
all checks have passed now

svettwer:
Hi!

Thx! I'll have a look as soon as possible. I'm currently very short on time.
Sorry for that.

BR,
Sven

Issue #583, opened 2019-01-04
CitrusFramework 2.7.6 - Param list support with same param name not available


rajivbansal2981:
**Citrus Version**:CitrusFramework 2.7.6

**Expected behavior**
In citrus framework till 2.7.5, multiple queryparam could be sent in HttpClientRequestActionBuilder and same behaviour should exist with the new version 2.7.6

**Actual behavior**
In 2.7.6, not able to send the array params as queryparam input in the request in HttpClientRequestActionBuilder

**Test case sample**
Looked the source code found that in 2.7.6

```java
public HttpMessage queryParam(String name, String value) {
            if (!StringUtils.hasText(name)) {
                throw new CitrusRuntimeException("Invalid query param name -         must not be empty!");
    }

    this.queryParams.put(name, value);

    String queryParamString = queryParams.entrySet()
                                         .stream()
                                         .map(entry -> entry.getKey() + (entry.getValue() != null ? "=" + entry.getValue() : ""))
                                         .collect(Collectors.joining(","));

    header(HttpMessageHeaders.HTTP_QUERY_PARAMS, queryParamString);
    header(DynamicEndpointUriResolver.QUERY_PARAM_HEADER_NAME, queryParamString);

    return this;
} 
```

where as in release prior to 2.7.6 it was
```java
public HttpMessage queryParam(String name, String value) {
            if (!StringUtils.hasText(name)) {
                throw new CitrusRuntimeException("Invalid query param name - must not be empty!");
    }

    String queryParams;
    if (getHeader(HttpMessageHeaders.HTTP_QUERY_PARAMS) != null) {
        queryParams = getHeader(HttpMessageHeaders.HTTP_QUERY_PARAMS).toString();
        queryParams += "," + name + (StringUtils.hasText(value) ? "=" + value : "");
    } else {
        queryParams = name + (StringUtils.hasText(value) ? "=" + value : "");
    }

    header(HttpMessageHeaders.HTTP_QUERY_PARAMS, queryParams);
    header(DynamicEndpointUriResolver.QUERY_PARAM_HEADER_NAME, queryParams);

    return this;
}
```
As a result if i use `queryParam("paramName", "value1").queryParam("paramName", "value2")` then only paramName with "Value1" is sent not "value2".

May I ask why this feature is removed or is there any workaround available?

svettwer:
Hi!

The change was made with #395. Unfortunately, I think this is a regression.

[...]

PR #634, opened 2019-03-27
Fix/583/multiple query params

cliffle:
#583 multiple queryParams with the same key are now possible

### DIFF
--- a/modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
+++ b/modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2014 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,8 +30,8 @@ import java.io.IOException;
 import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -223,14 +223,6 @@ public class HttpMessage extends DefaultMessage {
         return this;
     }
 
-    private void addQueryParam(final String name, final String value) {
-        if (this.queryParams.containsKey(name)) {
-            this.queryParams.get(name).add(value);
-        } else {
-            this.queryParams.put(name, new ArrayList<>(Collections.singletonList(value)));
-        }
-    }
-
     /**
      * Sets a new Http request query param.
      *
@@ -266,11 +258,7 @@ public class HttpMessage extends DefaultMessage {
         return this;
     }
 
-    private String outputQueryParam(Map.Entry<String, Collection<String>> entry) {
-        return entry.getValue().stream()
-                .map(entryValue -> entry.getKey() + (entryValue != null ? "=" + entryValue : ""))
-                .collect(Collectors.joining(","));
-    }
+
 
     /**
      * Sets request path that is dynamically added to base uri.
@@ -490,7 +478,7 @@ public class HttpMessage extends DefaultMessage {
     public void setCookies(final Cookie[] cookies) {
         this.cookies.clear();
         if (cookies != null) {
-            for (Cookie cookie : cookies) {
+            for (final Cookie cookie : cookies) {
                 cookie(cookie);
             }
         }
@@ -519,7 +507,7 @@ public class HttpMessage extends DefaultMessage {
      * @return The parsed dump as HttpMessage
      */
     public static HttpMessage fromRequestData(final String requestData) {
-        try (BufferedReader reader = new BufferedReader(new StringReader(requestData))) {
+        try (final BufferedReader reader = new BufferedReader(new StringReader(requestData))) {
             final HttpMessage request = new HttpMessage();
 
             final String[] requestLine = reader.readLine().split("\\s");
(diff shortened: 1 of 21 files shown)


modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
```diff
@@ -339,7 +344,7 @@ public String getAccept() {
      * Gets the Http request query params.
      * @return The query parameters as a key value map
      */
-    public Map<String, String> getQueryParams() {
+    public Multimap<String, String> getQueryParams() {
```
svettwer: This breaks the public API of `HttpMessage`. But in this case, we have to break the API so that the returned value is able to represent the query parameter content regarding its specification.
Nevertheless, I'd recommend to go wit a `Map<String, Collection<String>>` to keep the guava stuff internal and have an java standard conform interface to the outer world.

tools/test-generator/src/main/java/com/consol/citrus/generate/provider/http/ReceiveHttpRequestActionProvider.java
```diff
@@ -60,7 +61,7 @@ public ReceiveRequestModel getAction(String endpoint, HttpMessage message) {
 
         requestType.setHeaders(requestHeaders);
 
-        if (!CollectionUtils.isEmpty(message.getQueryParams())) {
+        if (!message.getQueryParams().isEmpty()) {
```
svettwer: This could cause a null pointer, if `message.getQueryParams()` returns `null`. If you change the signature of `message.getQueryParams()` back to a `Map<K, V>` implementation, `CollectionUtils.IsEmpty()` will be usable again, which includes a null check.

tools/test-generator/src/main/java/com/consol/citrus/generate/provider/http/ReceiveHttpRequestCodeProvider.java
```diff
@@ -56,7 +55,7 @@ public CodeBlock getCode(String endpoint, HttpMessage message) {
                     code.add(".header($S, $S)\n", entry.getKey(), Optional.ofNullable(entry.getValue()).map(Object::toString).orElse(""));
                 });
 
-        if (!CollectionUtils.isEmpty(message.getQueryParams())) {
+        if (!message.getQueryParams().isEmpty()) {
```
svettwer: This could cause a null pointer, if `message.getQueryParams()` returns `null`. If you change the signature of `message.getQueryParams()` back to a `Map<K, V>` implementation, `CollectionUtils.IsEmpty()` will be usable again, which includes a null check.

tools/test-generator/src/main/java/com/consol/citrus/generate/provider/http/SendHttpRequestActionProvider.java
```diff
@@ -60,7 +62,7 @@ public SendRequestModel getAction(String endpoint, HttpMessage message) {
 
         requestType.setHeaders(requestHeaders);
 
-        if (!CollectionUtils.isEmpty(message.getQueryParams())) {
+        if (!message.getQueryParams().isEmpty()) {
```
svettwer: This could cause a null pointer, if `message.getQueryParams()` returns `null`. If you change the signature of `message.getQueryParams()` back to a `Map<K, V>` implementation, `CollectionUtils.IsEmpty()` will be usable again, which includes a null check.

tools/test-generator/src/main/java/com/consol/citrus/generate/provider/http/SendHttpRequestCodeProvider.java
```diff
@@ -56,7 +55,7 @@ public CodeBlock getCode(String endpoint, HttpMessage message) {
                     code.add(".header($S, $S)\n", entry.getKey(), Optional.ofNullable(entry.getValue()).map(Object::toString).orElse(""));
                 });
 
-        if (!CollectionUtils.isEmpty(message.getQueryParams())) {
+        if (!message.getQueryParams().isEmpty()) {
```
svettwer: This could cause a null pointer, if `message.getQueryParams()` returns `null`. If you change the signature of `message.getQueryParams()` back to a `Map<K, V>` implementation, `CollectionUtils.IsEmpty()` will be usable again, which includes a null check.

svettwer:
We should also have a look at the duplicated code in `ReceiveHttpRequestCodeProvider`, `ReceiveHttpRequestActionProvider` and the respective `Send` counterpart. =)

modules/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java
```diff
@@ -195,9 +198,11 @@ public HttpMessage queryParams(String queryParamString) {
         header(HttpMessageHeaders.HTTP_QUERY_PARAMS, queryParamString);
         header(DynamicEndpointUriResolver.QUERY_PARAM_HEADER_NAME, queryParamString);
 
-        this.queryParams = Stream.of(queryParamString.split(",")).map(keyValue -> Optional.ofNullable(StringUtils.split(keyValue, "=")).orElse(new String[] {keyValue, ""}))
-                                                                        .filter(keyValue -> StringUtils.hasText(keyValue[0]))
-                                                                        .collect(Collectors.toMap(keyValue -> keyValue[0], keyValue -> keyValue[1]));
+        this.queryParams = Stream.of(queryParamString.split(","))
+                .map(keyValue -> Optional.ofNullable(StringUtils.split(keyValue, "=")).orElse(new String[] {keyValue, ""}))
+                .filter(keyValue -> StringUtils.hasText(keyValue[0]))
+                .collect(Multimaps.toMultimap(keyValue -> keyValue[0], keyValue -> keyValue[1], HashMultimap::create));
```
svettwer: We found out that the `Multimaps.toMultimap` collector is experimental in the current guava version. If we want to keep the Multimap in the project, we should make sure that the collector is stable in the newest guava version. Otherwise, we should switch back to a classical `Map<String, Collection<String>> implementation internally as well.

Issue #539, opened 2018-11-21
Support Cucumber Version >= 4.0.0


svettwer:
Extracted from #537
---------------

**User story**
As a Citrus user, I would like to use cucumber version >= 4.0.0 with my citrus tests, so that I can use the 4.x.x features of cucumber.

**Additional information**
- 2.7.x of citrus supports Cucumber 2.x
- 2.8.x of citrus supports Cucumber 3.x

svettwer:
Update: The major update to cucumber 4 is mostly done. I just have to add some docs and test before this is ready for review.

BR,
Sven

hmmlopez:
Hello Sven,
Will this branch eventually be merged with the master one? 
I'm looking forward to being able to use Cucumber 4 with Citrus. When merged with the master branch I could use it directly with the 2.9.0-SNAPSHOT until it's released (any info about the release of 2.9.0 maybe?).
For now, I'll use the 2.8.0 version with Cucumber 3, thanks for keeping up the good work with CitrusFramework.

svettwer:
Hi @hmmlopez,

yes, this branch will eventually be merged as soon as development and review have been finished. I just realized that in Cucumber 4, some API changes affect the Citrus after suite life cycle hooks. I am in touch with the Cucumber-jvm team to fix this but I have to spent some hours to get into cucumber a little bit more.

>When merged with the master branch I could use it directly with the 2.9.0-SNAPSHOT until it's released

Currently the master build does not publish a snapshot automatically at the moment, but I will create a snapshot after the merge now that I know it is requested. :+1: 

>any info about the release of 2.9.0 maybe

Yes, based on our [road map for 2019](http://citrusframework.org/news/2019/01/17/citrus-roadmap-2019/) 2.9.0 is _planned_ for Q2 so by the end of June. As I am currently working on a lot of customer projects, I am not sure whether I can guarantee a 2.9.0 with all the features.

>For now, I'll use the 2.8.0 version with Cucumber 3 [...]

Awesome! :tada: 

> [...] thanks for keeping up the good work with CitrusFramework. 

You're welcome :smiley: 

BR,
Sven

hmmlopez:
Hello Sven,

Thanks for the update. 
When merged with master, I could live with a local SNAPSHOT as well. But would be nice if I can pull the snapshot from the citrus repository directly, then I don't need to clone the repository and build it locally myself.

Thanks again,
Humphrey

hmmlopez:
[...]

PR #636, opened 2019-03-27
Small fix(crutch) to support cucumber4 gluePaths.

TimurSitdikov:
There is an [issue](https://github.com/citrusframework/citrus/issues/539) to add support for Cucumber>=4 . I've done a small fix for gluePaths type incompatibility. In Cucumber <= 4 gluePath is String, but in 4+ versions it's URI. So I've done crutch that just converts URI to String and Citrus starts normally in a bundle  with Cucumber 4+.


svettwer:
Hi!

I briefly checked the change and did a compile grade migration to cucumber jvm 4.2.5.
Unfortunately cucumber removed the public access to their reporter API but this is a different topic. I created an internal branch `maintenance/539/Support-Cucumber-4.x.x` for some further migration steps and cleanup. Thx a lot for boosting this! :+1: :tada: 

BR,
Sven

christophd:
We have updated to Cucumber 5 while going the path to Citrus 3.0. So this PR is obsolete

Issue #499, opened 2018-10-05
@ignore@ doesn't work for jdbc operations


laura098:
**Citrus Version** 2.7.6
**Expected behavior** using @ignore@ means that the data in a field will be ignored
**Actual behavior** the data is not ignored
**Test case sample**
In Citrus insert the testcase:
```xml
<receive endpoint="cxxDb">
	<message>
		<payload>
			<jdbc:operation>
				<jdbc:execute>
					<jdbc:statement>
						<jdbc:sql>INSERT INTO CXX.BXX (ID,ID_LXX,ZEITSTEMPEL,PRODUKT,PHASE) VALUES (CXX.SEQ_BXX.nextval,?,SYS_EXTRACT_UTC(CAST(? AS TIMESTAMP WITH TIME ZONE)),?,?) - (924,@ignore@,txx,40)</jdbc:sql>
					</jdbc:statement>
				</jdbc:execute>
			</jdbc:operation>
		</payload>
	</message>
</receive>
```
The Testcase has a timestamp. On each run of the test the timestamp has another value.  So the actual and the expected values are not equal.
Result:  
```
Validation failed: Node value not equal for element 'sql', expected 'INSERT INTO CCWEB.BTMDSPH (ID,ID_LXX,ZEITSTEMPEL,PRODUKT,PHASE) VALUES (CCODWA.SEQ_BXX.nextval,?,SYS_EXTRACT_UTC(CAST(? AS TIMESTAMP WITH TIME ZONE)),?,?) - (924,2018-10-01 16:53:38.561,txx,40)' but was 'INSERT INTO CXX.BXX(ID,ID_LXX,ZEITSTEMPEL,PRODUKT,PHASE) VALUES (CXX.SEQ_BXX.nextval,?,SYS_EXTRACT_UTC(CAST(? AS TIMESTAMP WITH TIME ZONE)),?,?) - (924,2018-10-01 16:57:46.172,txx,40)'
```
This problem would be solved, if citrus can ignore the timestamp.
 Actual result if insert ignore method on timestamp field:
```
[...]

PR #643, opened 2019-04-24
Enhancement/499/add plaintext validation to sql statements

svettwer:
PR concerning #499

### DIFF
--- a/modules/citrus-core/src/main/java/com/consol/citrus/validation/json/JsonTextMessageValidator.java
+++ b/modules/citrus-core/src/main/java/com/consol/citrus/validation/json/JsonTextMessageValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2011 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


tnobody:
So as far as I understand the @ignore()@ pattern works somehow like `.*` in a regular expression, right?

So for my understanding it looks good.

Tests also looks good to me, you could consider to add negative Tests for ValidationMatcherUtils.java. 
Something like: 
```java
        final String controlMessage = "(924,@ignore(22)@,txx,40)";
        final String receivedMessage = "(924,2018-10-01 16:53:38.561,txx,40)";
```
Which should fail.

One proposal (which is not directly part of this PR/Issue):
This @ignore()@ validation-rule is now directly/hardcoded into the Validators (PlainText and Json) would it make sense to create API to generically add such rules to an Validator-class?

svettwer:
Hi!

Thx for the review! Yes a general validator class would make perfect sense to me but currently there is no requirement for such an refactoring/API.

Yes, a `@ignore@` statement works similar to an `.*` in a regex. A regex would follow some other rules in detail.

The idea of a negative test is also great. I'll add one and merge the branch after that.

BR,
Sven

PR #648, opened 2019-05-06
Suppress exceptions when XPath evaluations return false

tvelkoff:
Citrus currently reacts to failed XPath assertions by throwing exceptions. This pull request allows these to be treated as failures without disrupting the normal control flow of each test. There are no breaking changes to the API, although the behavior is necessarily changed. But the new behavior is optional and depends on a configuration property whose default maintains the current behavior of Citrus.

Therefore, this pull request maintains backward compatibility for existing Citrus tests, but offers a path forward for testers that choose to follow this approach to XPath evaluation.

The behavior is selected by the boolean property citrus.core.validation.xml.suppress_exceptions_on_xpath_evaluations. The default is false, which preserves the current behavior of Citrus. Setting the property to true enables the behavior changes implemented in this pull request.

The TestContext class maintains a list of exceptions thrown in the course of a test. Here, a parallel structure has been added to maintain a list of failures. Failures are described by a new immutable class, XpathAssertionResult. The finish method of TestCase sets the member variable testResult by examining the TestContext. A new method (evaluateTestResult) is called in order to set the value of testResult, setting it to failed if failures have been detected and the above property is true.

The XpathMessageValidator's validateMessage method validates a set of XPath expressions, throwing a ValidationException if any of them fail. In this pull request, this behavior is altered by the validateValues method, which is called from validateMessage. This method catches ValidationExceptions, and if the above property is true, it suppresses the exceptions and adds an XpathAssertionResult to the list of failures in the TestContext. If the property is false, it re-throws the ValidationException, which preserves the current behavior of Citrus.

The method extractVariables of class XpathPayloadVariableExtractor evaluates the XPath expressions in the TestContext against the current Message, and stores the result of each evaluation as a variable in the TestContext. If the evaluation finds no result, it throws a CitrusRuntimeException. This pull request alters this behavior (as usual, controlled by the property, and preserving the current behavior when it is false) by catching CitrusRuntimeExceptions and producing a value of an empty string. The use of the empty string is significant because this result becomes the value of a Citrus variable, which cannot be null.

### DIFF
--- a/modules/citrus-core/src/main/java/com/consol/citrus/TestCase.java
+++ b/modules/citrus-core/src/main/java/com/consol/citrus/TestCase.java
@@ -18,6 +18,7 @@ package com.consol.citrus;
 
 import com.consol.citrus.container.*;
 import com.consol.citrus.context.TestContext;
+import com.consol.citrus.context.XpathAssertionResult;
 import com.consol.citrus.exceptions.CitrusRuntimeException;
 import com.consol.citrus.exceptions.TestCaseFailedException;
 import com.consol.citrus.report.TestActionListeners;
@@ -81,6 +82,12 @@ public class TestCase extends AbstractActionContainer implements BeanNameAware {
 
     /** Logger */
     private static Logger log = LoggerFactory.getLogger(TestCase.class);
+
+    private static boolean suppressExceptionsOnXpathEvaluations;
+    
+    static {
+    	suppressExceptionsOnXpathEvaluations = Boolean.parseBoolean(System.getProperty("citrus.core.validation.xml.suppress_exceptions_on_xpath_evaluations", "false"));
+    }
 
     /**
      * Starts the test case.
@@ -273,6 +280,7 @@ public class TestCase extends AbstractActionContainer implements BeanNameAware {
             if (contextException != null) {
                 throw new TestCaseFailedException(contextException);
             }
+            testResult = evaluateTestResult(testResult, context);
         } catch (final TestCaseFailedException e) {
             throw e;
         } catch (final Exception | AssertionError e) {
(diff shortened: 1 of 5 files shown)


tvelkoff:
Sorry, I haven't used TestNG before and I need to get this up in Eclipse. I will resubmit after ensuring the tests pass.

christophd:
I wonder why a test should continue when a XPath validation failed to evaluate? Isn't that a critical assertion that should fail the test.

Also I think a test failing to extract values using XPath will not be able to continue as the extracted value most likely will be used in one of the next steps

christophd:
I think the introduced property changes the behavior of the whole test suite which is not the best solution in my point of view.

Instead we should have a function or validation matcher that explicitly catches errors and provides a compensation value. This way the user can explicitly change the behavior of individual XPath evaluations

tvelkoff:
Christoph, thank you for your comments. I'm aware that I'm touching on a fundamental design decision behind Citrus, so I tread carefully... :) Permit me to describe the use case behind this suggestion.

I have created a Spring Boot application that uses a custom JUnit 5 launcher to run system tests where I use the Citrus framework to call the SUT using a web service and a queue of results. I use Citrus to perform XPath evaluations of the results. Ultimately the test results are written to Postgres and displayed using Grafana. I have dozens of such tests and run the test suite daily. This results in time series data, which permits trends to develop for each test case over time. The (non-production) SUT interacts with many other (non-production) systems, and it is not uncommon that any of them could be down at the time the tests run. Two other common situations are 1) the data in the external system changed - which means the expected test results need to be updated or 2) the expected values in the test are incorrect to begin with and need to be fixed. Over time, the quality of the tests will improve, with the goal that the only remaining failing tests will be those due to timeouts.

So... it is normal for a test to fail, for example, if the XPath assertion expects a count of 1 but is actually zero. In this case, I certainly _do_ want to record this as a failure, but would prefer not to have the evaluation cause an exception. That's the chief motivation behind this proposal.

I also collect diagnostics about each test that can be used to correct them (i.e. change the expected values). Some of these diagnostics are values found in the result message, but are optional. In fact, they are typically present when there is a failure identified by the SUT itself. This is the motivation behind the need to tolerate empty XPath evaluations. This usually occurs only when the test is actually passing, and an exception would generate a false positive.

Thanks for bearing with me here, as I know this was a long post! I look forward to your further comments.

I will address the issues surrounding the property in a separate post.

tvelkoff:
Christoph, again thanks for your comments. Let me discuss the suggested property here.

While I was careful in this proposal not to make a breaking change in the API, necessarily I am introducing a change to behavior. I knew that if I simply proposed a change in the behavior, that this would invalidate all existing Citrus tests - certainly quite unacceptable!

The use of the property was intended to ensure that existing tests would continue to exhibit the same behavior as today. The default value ensures that this is so. In order to obtain the proposed behavior change, the tester has to explicitly call for it by setting the property.

I am by no means wedded to the use of a property to preserve the existing behavior of Citrus. I would be happy to learn how to accomplish this in a manner that is idiomatic in Citrus, as you suggested.

tvelkoff:
Of course there is a third reason for a failure of an XPath assertion - so obvious I forgot about it! The test in fact could be correct and the SUT has produced an incorrect result.

tvelkoff:
I'm closing this pull request in favor of submitting a different pull request with a (hopefully) better solution.
I agree that a global property is a blunt instrument to be avoided. Instead, I plan to introduce an 'isLenient" flag in XpathMessageValidator and XpathPayloadVariableExtractor. Instances of these classes can be injected into the ReceiveMessageAction using the Builder. The Builder currently supports the validator and it looks like an extractor method will need to be added.
I expect to submit a replacement pull request in the next week or so.

PR #776, opened 2021-06-02
(#775) Altered setter on SleepActionFactoryBean

### LINKED ISSUE #775: Sleep with Variable fails in XML DSL
**Citrus Version**
3.0.0-M3

**Expected behavior**
The following expression should sleep as long as `${timeToSleep}` defines
```xml
<sleep time="${timeToSleep}" />
```

**Actual behavior**
```
com.consol.citrus.exceptions.CitrusRuntimeException: Failed to load test case
[...]
Initialization of bean failed; nested exception is org.springframework.beans.TypeMismatchException: Failed to convert property value of type 'java.lang.String' to required type 'java.lang.Double' for property 'seconds'; nested exception is java.lang.NumberFormatException: For input string: "${timeToSleep}"
```

**Test case sample**
```xml
<spring:beans xmlns="http://www.citrusframework.org/schema/testcase"
        xmlns:spring="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
          http://www.citrusframework.org/schema/testcase http://www.citrusframework.org/schema/testcase/citrus-testcase.xsd">

  <testcase name="SleepWithVariable">

    <variables>
      <variable name="timeToSleep" value="5"/>
    </variables>

    <actions>
      <sleep time="${timeToSleep}" />
    </actions>
  </testcase>
</spring:beans>
```

svettwer:
Resolves #775

### DIFF
--- a/core/citrus-spring/src/main/java/com/consol/citrus/config/xml/SleepActionParser.java
+++ b/core/citrus-spring/src/main/java/com/consol/citrus/config/xml/SleepActionParser.java
@@ -24,6 +24,8 @@ import org.springframework.beans.factory.xml.BeanDefinitionParser;
 import org.springframework.beans.factory.xml.ParserContext;
 import org.w3c.dom.Element;
 
+import java.util.concurrent.TimeUnit;
+
 /**
  * Bean definition parser for sleep action in test case.
  *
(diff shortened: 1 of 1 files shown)


christophd (approved):
Thanks!

PR #777, opened 2021-06-07
(#644) Don't clear cache after accessing it.

### LINKED ISSUE #644: MessageTracingTestListener changed output from citrus v2.7.5
**Citrus Version**
since v2.7.5

**Expected behavior**
Using an extended MessageTracingTestListener the method onOutboundMessage should have the headers and the body of the message.

**Actual behavior**
The message of onOutboundMessage contains only the headers, no body.
E.g. it shows a `Content-Length: 65` but no content following.

**Additional information**
Using Citrus v2.7.4 it works as expected without any other changes.

**Test case sample**
To test simply send a http response from Citrus and take a look in the log.

svettwer:
Fixes #644 

As mentioned in #644, changes in #359 lead to a changed behavior when accessing message logs via `MessageTracingTestListener`. This is because in 802f7b5, a Queue was introduced that polls its elements on access. 
Poll is described as `Retrieves and removes the head of this queue, or returns null if this queue is empty`. If more than one implementation of `LoggingHandlerInterceptor` is registered, the queue is accessed multiple times but only the first call to the queue will receive the communication due to the implementation of poll.

I see the advantages of adding a ConcurrentLinkedQueue here but with the current implementation this causes a regression I'd like to avoid.

### DIFF
--- a/endpoints/citrus-http/src/main/java/com/consol/citrus/http/controller/HttpMessageController.java
+++ b/endpoints/citrus-http/src/main/java/com/consol/citrus/http/controller/HttpMessageController.java
@@ -21,18 +21,24 @@ import com.consol.citrus.endpoint.adapter.EmptyResponseEndpointAdapter;
 import com.consol.citrus.http.client.HttpEndpointConfiguration;
 import com.consol.citrus.http.message.HttpMessage;
 import com.consol.citrus.message.Message;
-import org.springframework.http.*;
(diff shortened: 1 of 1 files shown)


christophd:
@svettwer I have another PR that should fix the issue please have a look #782 

This should fix the regression introduced when multiple handler interceptors try to get the very same response.

svettwer:
closed in favor of #782

PR #782, opened 2021-07-13
fix(#644): Cache server responses according to servlet request

### LINKED ISSUE #644: MessageTracingTestListener changed output from citrus v2.7.5

christophd:
- Introduce map cache with servlet request as key and response as value
- Access cache with servlet request to make sure that multiple handler interceptors get the very same response content
- Introduce cache size to avoid memory issues with long running server instances
- Automatically clear cache when maximum number of responses is reached

This fixes #644 with multiple logging interceptor handlers calling for the response content. All interceptors access the cache with the same servlet request. This makes sure that each interceptor gets access to the very same response content.

The downside is that the in memory cache will grow with every communication on the server. Therefore a maximum cache size has been introduced and a mechanism to automatically free the cache when it is full.

### DIFF
--- a/endpoints/citrus-http/src/main/java/com/consol/citrus/http/config/annotation/HttpServerConfig.java
+++ b/endpoints/citrus-http/src/main/java/com/consol/citrus/http/config/annotation/HttpServerConfig.java
@@ -136,6 +136,12 @@ public @interface HttpServerConfig {
     HttpStatus defaultStatus() default HttpStatus.OK;
 
     /**
+     * Server default response cache size.
+     * @return
+     */
+    int responseCacheSize() default 100;
+
+    /**
      * Binary media types.
      * @return
      */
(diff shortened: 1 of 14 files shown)


svettwer (approved):
LGTM! :+1: 
I can approve that #644 is fixed with this changes in a local setup. What I like about the solution is the cache implementation where the message history is preserved and the memory is protected from overflow,

PR #812, opened 2021-10-25
fix(core): jdk17 compatibility of citrus

bbortt:
update dependencies and fix tests for JDK17.


pom.xml
```diff
@@ -178,7 +176,7 @@
     <htmlunit.version>2.49.1</htmlunit.version>
     <byte.buddy.version>1.11.0</byte.buddy.version>
     <google.guava.version>30.1.1-jre</google.guava.version>
-    <surefire.version>2.22.2</surefire.version>
+    <failsafe.surefire.version>2.22.2</failsafe.surefire.version>
```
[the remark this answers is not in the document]
christophd: wonder if it makes sense to introduce surefire and failsafe version so these two can diverge in future

pom.xml
```diff
@@ -136,13 +136,11 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <maven.compiler.release>11</maven.compiler.release>
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
```
christophd, line 141: using `maven.compiler.source` and `maven.compiler.target` properties is kind of standard practice. it has the effect to overwrite the Maven compiler plugin properties. I used to know some occasions where this is safer than setting the values in the plugin configuration with a custom property (e.g. java.version).

I think it has to do something with using `-Dmaven.compiler.target=14` which is commonly known by Maven users I guess.

christophd (commented):
Looking good! Many thanks!

Some comments and nitpicks as usual

pom.xml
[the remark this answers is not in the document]
bbortt: never seen these two diverge. neither seen these two being set in one property :D I think I'll revert that.

pom.xml
```diff
@@ -1143,6 +1141,36 @@
...
+        <groupId>org.apache.maven</groupId>
+        <artifactId>maven-core</artifactId>
+        <scope>provided</scope>
+        <version>3.6.1</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.maven</groupId>
+        <artifactId>maven-model</artifactId>
+        <scope>provided</scope>
+        <version>3.8.3</version>
```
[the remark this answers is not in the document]
bbortt: actually, I wanted to update the maven wrapper (is on `3.6.3`) and push `3.6.1` to `3.8.3`. looks like I stopped halfway through, cause some dependencies are still on `3.6.1`. I'll align all of those to `3.8.3`.

pom.xml
bbortt: I can accept that. will change that to:
```
    <java.version>11</java.version>
    <maven.compiler.release>${java.version}</maven.compiler.release>
    <maven.compiler.source>${java.version}</maven.compiler.source>
    <maven.compiler.target>${java.version}</maven.compiler.target>
```

bbortt:
> Some comments and nitpicks as usual

no worries, those are good inputs! should have caught most of 'em 🚀

pom.xml
```diff
@@ -1229,11 +1262,11 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>3.8.1</version>
+          <version>${maven.version}</version>
```
christophd, line 1265: The plugin version does not always comply with the Maven version. It is too dangerous and might lead to future build failures

pom.xml
```diff
@@ -1258,7 +1291,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-plugin-plugin</artifactId>
-          <version>3.5</version>
```
christophd, line 1261: The plugin version does not always comply with the Maven version. It is too dangerous and might lead to future build failures

tools/archetypes/jms/src/main/resources/archetype-resources/pom.xml
```diff
@@ -89,17 +89,17 @@
 
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
-        <version>3.6.1</version>
+        <version>${maven.version}</version>
```
christophd, line 92: we need to add `maven.version` and `java.version` properties in this POM file. The properties of the parent POM will not be available at runtime in the archetpye

core/citrus-spring/src/test/resources/com/consol/citrus/config/xml/SequenceAfterSuiteParserTest-context.xml
```diff
@@ -30,9 +30,6 @@
   </citrus:after-suite>
 
   <citrus:after-suite id="afterSuite3">
-    <citrus:env>
-      <citrus:property name="USER"/>
```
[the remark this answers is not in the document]
christophd: How about using PATH environment variable instead? This seems to be a valid env setting on multiple OS (unix, MacOS, windows, ...)

christophd (commented):
Awesome! Some comments to be discussed though

pom.xml
bbortt: one could overwrite/force another version if it leads to failures on his site.. or do you mean inside citrusframework/citrus?

core/citrus-spring/src/test/resources/com/consol/citrus/config/xml/SequenceAfterSuiteParserTest-context.xml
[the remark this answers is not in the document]
bbortt: that is a good idea.. we could even a windows-action later, maybe? let me dig into this.

tools/archetypes/jms/src/main/resources/archetype-resources/pom.xml
bbortt: I didn't know that. will revert those changes.

pom.xml
christophd: The maven compiler plugin is not tied to the Maven version. The versions may differ so using the `${maven.version}` property here is not what we should do. Either stick with hardcoded version 3.8.1 or introduce another property like ${maven.compiler.version} and give it the value 3.8.1

pom.xml
christophd: @bbortt I think this is why the build_jdk11 job is failing

pom.xml
bbortt: done 👍🏼

bbortt:
**this pull-request has label: `WIP`!**

bbortt:
@christophd do you have any idea on the current build failure? it looks good on localhost!

bbortt:
@christophd you can remove the label: `WIP`. it's ready for review. I think `com.consol.citrus` now really is jdk 17 compatible 🥂 
btw. are you gonna do the renaming to `org.citrusframework`, or should I?

christophd:
@bbortt that's fantastic news! many thanks!

I have started the build once again.

Regarding the renaming to `org.citrusframework` I think we should start working on that on a new 4.x branch. I guess we need to do this with a major release. Let me create the branch

christophd:
@bbortt had to release v3.1.2 due to log4j vulnerability. Could you please rebase the branch and resolve the conflicts?

bbortt:
> Could you please rebase the branch and resolve the conflicts?

@christophd done. so, we had five builds and five different failures 😲 interesting.. seems to be trouble with the JDK17 build in some way, because the JDK11 build is green as a 🍀 I've summarized that in #835.

Issue #822, opened 2021-12-09
Validation of request


tschlat:
**User story**
As a test designer I want to be able to validate my outgoing request. This ensures that my test case always sends valid requests. Validation could be performed against JsonSchema or xsd. 

Currently Citrus only supports the validation of the response.

christophd:
+1

tschlat:
Has been implemented and merged into master.

https://github.com/citrusframework/citrus/commit/c9e3d5adc23e7a8dfe54bd0fb9d0a66a4a778d4b

PR #839, opened 2022-01-27
fix(#822): Schemavalidation of message on send

tschlat:
Implement schema validation on send message for XML and JSON as described in #822.
 
@christophd please note the TODO in JsonTextMessageValidator


christophd:
@tschlat many thanks for the initiative. Having a look at your changes I wonder if using the MessageValidator as a whole for this is too heavy. The schema validation capabilities is only a small subset of the message validator. How about just using the schema validators directly for send operations instead?

I think we do not need the whole validation context and message validator which adds a huge amount of methods to the Java DSL. Instead I would try to apply the schema validation in send directly

WDYT?

christophd:
The message validator only delegates work to the schema validation:
https://github.com/citrusframework/citrus/blob/main/validation/citrus-validation-json/src/main/java/com/consol/citrus/validation/json/schema/JsonSchemaValidation.java

https://github.com/citrusframework/citrus/blob/main/validation/citrus-validation-json/src/main/java/com/consol/citrus/validation/json/JsonTextMessageValidator.java#L134

My idea is to only use this part in send operation directly. so we do not have to add handling of validation contexts and other stuff to the send operation.

christophd:
of course we can have options to set schema or schemaRepository and to enable/disable schema validation on the send operation. this would only add few more options to the Java DSL builder

tschlat:
I needed to introduce a new flag on MessageValidator. "supportsSendMessage". This is necessary to avoid that PlainTextMessageValidators used by several tests kick in and try to validate the message. Note that the send message currently has no control message. The send message validation is meant to perform only schema validation to make sure that API changes (in wsdl/xsd/json) can be detected on send message. Any other ideas on how to solve  this are appreciated.

> @tschlat many thanks for the initiative. Having a look at your changes I wonder if using the MessageValidator as a whole for this is too heavy. The schema validation capabilities is only a small subset of the message validator. How about just using the schema validators directly for send operations instead?
> 
> I think we do not need the whole validation context and message validator which adds a huge amount of methods to the send operation in Java DSL. Instead I would try to apply the schema validation in send directly
> 
> WDYT?

I dont't see how this could possibly be done as the SendMessageAction is not aware of any Message type. And there is nothing like a SchemaValidator available. As you have probably seen I added the following attributes to SendMessage Schema:
- schemaValidation (which activates schema validation in general)
- schema (to specify the schema to validate against)
- schem-repository (to specify the schema-repository to use for validation)

In addition I would like to be able to activate the validation globally. Therefore the new system property you find in SendMessageAction.

Could you please elaborate on your idea so that I can move the implementation in that direction.

christophd:
so my idea was to **not** introduce message validators to send operation at all because this brings a lot of configuration and heavy weight to the operation. 

Instead add schema validation related settings like you did with 
* schemaValidation (which activates schema validation in general)
* schema (to specify the schema to validate against)
* schem-repository (to specify the schema-repository to use for validation)
and then try to invoke schema validation directly from send operation.

The missing message type on send operation is a challenge. If schema validation is enabled I would auto detect message type from given send message payload. There are utility methods to check for `isJson()` and `isXml()` on a message payload. Once we have this we can do a resource lookup for a schema validator for the given message type. The schema validator is something like [JsonSchemaValidation](https://github.com/citrusframework/citrus/blob/main/validation/citrus-validation-json/src/main/java/com/consol/citrus/validation/json/schema/JsonSchemaValidation.java). The validation is capable of performing a schema validation on a payload. The send operation can do the lookup and perform the validation if enabled.

Let me know if this works for you.

christophd:
> In addition I would like to be able to activate the validation globally. Therefore the new system property you find in SendMessageAction.

Activating the validation globally with system properties is a good idea! I would add these settings to [CitrusSettings](https://github.com/citrusframework/citrus/blob/main/core/citrus-api/src/main/java/com/consol/citrus/CitrusSettings.java) class though and also support environment variables when setting these options.

tschlat:
> so my idea was to **not** introduce message validators to send operation at all because this brings a lot of configuration and heavy weight to the operation.
> 
> Instead add schema validation related settings like you did with
> 
> * schemaValidation (which activates schema validation in general)
> * schema (to specify the schema to validate against)
> * schem-repository (to specify the schema-repository to use for validation)
>   and then try to invoke schema validation directly from send operation.
> 
> The missing message type on send operation is a challenge. If schema validation is enabled I would auto detect message type from given send message payload. There are utility methods to check for `isJson()` and `isXml()` on a message payload. Once we have this we can do a resource lookup for a schema validator for the given message type. The schema validator is something like [JsonSchemaValidation](https://github.com/citrusframework/citrus/blob/main/validation/citrus-validation-json/src/main/java/com/consol/citrus/validation/json/schema/JsonSchemaValidation.java). The validation is capable of performing a schema validation on a payload. The send operation can do the lookup and perform the validation if enabled.
> 
> Let me know if this works for you.

Fine with me. Before I implement this, let me please sketch the necessary steps to align with your thoughts. This is what I would do to accomplish the task:

1. JSON and XML Schema validation need the ValidationContext thus the DSL for SendMessage will not really change as I need to be able to add the validation context which is not really different from now. As an alternative I could simply collect the properties for "schemaValidation", "schema" and "schema-repository" and create a proper validation context on the fly when I need in SendMessage.
2. The is no common interface for SchemaValidation. In order to retrieve a message specific SchemaValidation I need to introduce this interface and let JsonMessageValidation implement it. Furthermore I need to extract the XmlMessageValidation from XmlDomMessageValidator and move all schema specifc validation into it.
3. I need to register the SchemaValidators similar to MessageValidators

This approach would clearly separate the SchemaValidation concerns from the MessageValidation concerns. XML/Json MessageValidators could simply delegate the schema validation to respective SchemaValidators. 
In the end I could request the message specific SchemaValidators from SendMessageAction.validateMessage() and use them to only validate the schema aspect of the message.

Please let me know if this is what you are aiming for.

christophd:
@tschlat that's what I have had in mind. Great summary and exactly what we need. Many thanks!

> As an alternative I could simply collect the properties for "schemaValidation", "schema" and "schema-repository" and create a proper validation context on the fly when I need in SendMessage.

I think this would be a good solution and we can keep the Java DSL for send operation a bit more straight forward because we do not have to handle the validation context (with reconcile loops and so on).

many thanks again, looking forward to your solution

tschlat:
I have implemented as discussed. For the sake of simplicity I have used the existing MessageValidatorRegistry to also manage the SchemaValidators.

tschlat:
@christophd I have a green build locally on windows. Do you have an idea whats going on with this JMS?

endpoints/citrus-ws/src/main/resources/com/consol/citrus/schema/citrus-testcase.xsd
```diff
@@ -460,6 +463,36 @@
         </xs:sequence>
     </xs:complexType>
 
+    <xs:complexType name="PurgeJmsQueuesActionType">
```
christophd, line 466: where does this come from?

endpoints/citrus-ws/src/main/resources/com/consol/citrus/schema/citrus-testcase.xsd
```diff
@@ -876,6 +909,8 @@
             <xs:element ref="echo"/>
             <xs:element ref="expect-timeout"/>
             <xs:element ref="purge-endpoint"/>
+            <xs:element ref="purge-jms-queues"/>
```
christophd, line 912: why is that being added?

endpoints/citrus-ws/src/main/resources/com/consol/citrus/schema/citrus-testcase.xsd
```diff
@@ -945,6 +980,8 @@
     <xs:element name="echo" type="EchoActionType"/>
     <xs:element name="trace-variables" type="TraceVariablesActionType"/>
     <xs:element name="purge-endpoint" type="PurgeEndpointActionType"/>
+    <xs:element name="purge-jms-queues" type="PurgeJmsQueuesActionType"/>
+    <xs:element name="purge-channel" type="PurgeChannelActionType"/>
```
christophd, line 984: same as above, why added?

validation/citrus-validation-json/src/main/java/com/consol/citrus/validation/json/JsonTextMessageValidator.java
```diff
@@ -71,15 +71,19 @@
...
-        }
 
         log.debug("Start JSON message validation ...");
 
         if (validationContext.isSchemaValidationEnabled()) {
-            performSchemaValidation(receivedMessage, context, validationContext);
+            jsonSchemaValidation.validate(receivedMessage, context, validationContext);
+        }
+
+        // TODO: @Christoph do you have any doubts about moving this check behind the schema validation?
```
christophd, line 81: Can we revert this change? Now with the new solution where the schema validator performs the validation on send operation instead of the message validator?

christophd: for backward compatibility reasons. we can do this with the next major release though

validation/citrus-validation-json/src/main/resources/META-INF/citrus/message/schemaValidator/jsonSchema
```diff
@@ -0,0 +1,2 @@
+name=defaultJsonSchemaValidator
```
christophd, line 1: I wonder weather the file should be named `json` instead of `jsonSchema` so someone can lookup the validator by its name easily. Also then someone could get the default validator for the message type `json` by name. WDYT?

validation/citrus-validation-xml/src/main/resources/META-INF/citrus/message/schemaValidator/xmlSchema
```diff
@@ -0,0 +1,2 @@
+name=defaultXmlSchemaValidator
```
christophd, line 1: same comment as above. consider naming the file `xml` instead of `xmlSchema`

christophd (commented):
Wow, fantastic work! Many thanks. Left some nitpicker comments

christophd:
@tschlat Unfortunately JMS test is a flaky one from time to time

PR #844, opened 2022-02-01
Issue/818/xml test loader configuration

tschlat:
I have added an annotation XmlTestLoaderConfiguration that can be used to configure the XmlTestLoader to create specific types of TestCase and TestCaseMetaInfo instances.

The main idea is to replace the TestCaseParser in the CitrusTestcaseNamespaceHandler with a specifc implementation. To do so I added the new annotation that provides the replacement parser class. This parser is then injected into the CitrusTestcaseNamespaceHandler. The replacement magic happens in XmlTestLoader.


endpoints/citrus-spring-integration/src/test/java/com/consol/citrus/integration/common/XmlTestLoaderIT.java
```diff
@@ -0,0 +1,81 @@
...
+    @Test
+    @CitrusXmlTest
+    public void XmlTestLoaderIT() {
+
+        TestCase testCase = getTestCase();
+        Assert.assertTrue(testCase instanceof SampleTestCase);
+
+        TestCaseMetaInfo metaInfo = testCase.getMetaInfo();
+        Assert.assertTrue(metaInfo instanceof SampleTestCaseMetaInfo);
+        Assert.assertEquals(((SampleTestCaseMetaInfo)metaInfo).getDescription(), "Foo bar: F#!$§ed up beyond all repair");
```
bbortt, line 31: 🤣

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
```diff
@@ -91,7 +98,25 @@ private ApplicationContext loadApplicationContext() {
...
                             "com/consol/citrus/spring/annotation-config-ctx.xml"},
-                    true, getParentApplicationContext());
+                    true, getParentApplicationContext()) {
+
+                /**
+                 * Configures the namespace handler resolver with the {@link XmlTestLoaderConfiguration}
+                 * @param reader
+                 */
+                @Override
+                protected void initBeanDefinitionReader(XmlBeanDefinitionReader reader) {
```
bbortt, line 108: id probably create a `CitrusClassPathXmlApplicationContext` just for the sake of readability..

core/citrus-spring/src/main/java/com/consol/citrus/config/handler/CitrusTestcaseNamespaceHandler.java
```diff
@@ -28,7 +29,7 @@
  * Namespace handler registers bean definition parser
  * for Citrus testcase schema elements.
  *
- * @author Christoph Deppisch
+ * @author Christoph Deppisch, Thorsten Schlathoelter
  * @since 2007
  */
 public class CitrusTestcaseNamespaceHandler extends NamespaceHandlerSupport {
```
bbortt, line 35: I'd rename this class as `TestCase` is commonly written in camel-case. if @christophd thinks that this would not break anything?

```suggestion
public class CitrusTestCaseNamespaceHandler extends NamespaceHandlerSupport {
```

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
```diff
@@ -91,7 +98,25 @@ private ApplicationContext loadApplicationContext() {
...
+
+                    XmlTestLoaderConfiguration loaderConfiguration = testClass.getAnnotation(XmlTestLoaderConfiguration.class);
+                    if (loaderConfiguration != null) {
+                        reader.setNamespaceHandlerResolver(new TestLoaderConfigurationApplicator(loaderConfiguration, reader.getNamespaceHandlerResolver()));
+                    }
+                    super.initBeanDefinitionReader(reader);
```
bbortt, line 114: I think `super`-invocation should stand at the beginning, no? because it would possibly overwrite `reader#setNamespaceHandlerResolver` again.. altough, in this case it does most probably only invoke `reader#setValidating`..

bbortt (changes requested):
I see how this works, very nice @tschlat!

apart from the commented things I'd probably remove some trailing newlines 😉

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
tschlat: This is not possible as loaderConfig is referenced in the overloaded initBeanDefinitionReader() AND initBeanDefinitionReader() is called from the constructor of the application context. Thus there is no simple way to pass in the required loaderConfig - at least I don't see an other solution than creation of anonymous class here.

core/citrus-spring/src/main/java/com/consol/citrus/config/handler/CitrusTestcaseNamespaceHandler.java
tschlat: I have my doubts about this, as this handler is reflectively created from spring.handlers.
Of course I can adjust it but I don't know if there is a use case where a user has explicitly configured a spring.handlers specifying this class. This would of course be a breaking change.

I leave it for @christophd to decide whether this cosmetic change is worth it.

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
tschlat: If we take ClassPathXmlApplicationContext as a blackbox I would certainly wrap the resolver into our own implementation before passing it into super. Therefore the solution to call super at the end was explicitly chosen and actually I would want to stick with it.

core/citrus-spring/src/main/java/com/consol/citrus/config/handler/CitrusTestcaseNamespaceHandler.java
christophd: We should be fine to change the name as well as in spring.handlers in our the code base. The chance of someone referencing the old name in a custom spring.handlers is quite low. And if so the change is quite simple.

so ok with me

christophd (commented):
Love it! Great enhancement!

Only thing I would like to discuss is the fact that the user will have to annotate all of the tests to activate the custom parser. How about adding another option to overwrite the default parser via resource path lookup?

Something like this is already done for action parsers:
https://github.com/citrusframework/citrus/blob/main/core/citrus-spring/src/main/java/com/consol/citrus/config/TestActionRegistry.java

Users can add a custom parser with resource path lookup. We could search for a given custom parser via resource path and fallback to the default parser if non is set. This way the user has to provide the parser only once per project.

WDYT?

tschlat:
> Love it! Great enhancement!
> 
> Only thing I would like to discuss is the fact that the user will have to annotate all of the tests to activate the custom parser. How about adding another option to overwrite the default parser via resource path lookup?
> 
> Something like this is already done for action parsers: https://github.com/citrusframework/citrus/blob/main/core/citrus-spring/src/main/java/com/consol/citrus/config/TestActionRegistry.java
> 
> Users can add a custom parser with resource path lookup. We could search for a given custom parser via resource path and fallback to the default parser if non is set. This way the user has to provide the parser only once per project.
> 
> WDYT?

I had that in mind but forgot about it. I will add that configuration option  over the weekend.

tschlat:
> > Love it! Great enhancement!
> > Only thing I would like to discuss is the fact that the user will have to annotate all of the tests to activate the custom parser. How about adding another option to overwrite the default parser via resource path lookup?
> > Something like this is already done for action parsers: https://github.com/citrusframework/citrus/blob/main/core/citrus-spring/src/main/java/com/consol/citrus/config/TestActionRegistry.java
> > Users can add a custom parser with resource path lookup. We could search for a given custom parser via resource path and fallback to the default parser if non is set. This way the user has to provide the parser only once per project.
> > WDYT?
> 
> I had that in mind but forgot about it. I will add that configuration option over the weekend.

I had a look into it and realized, that I did not notice the existence of the TestActionRegistry. Thanks to @christophd for this link. 
Having this static registry makes things more simple and thus I reworked my implementation. Please notice that I have renamed the TestActionRegistry into CitrusNamespaceParserRegistry because this name better reflects the fact that it is now also used to override the TestCase parser and the TestaCaseMetaInfo parser. Due to this renaming  there are a couple of more changed files in the MR. 
I have changed the XmlTestLoaderConfiguration annotation to now allow overriding any of the TestCaseNamespaceParser parsers.

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
```diff
@@ -98,6 +111,26 @@ private ApplicationContext loadApplicationContext() {
         }
     }
 
+    /**
+     * Configures the CitrusNamespaceParserRegistry with custom parsers
+     */
+    private void configureCustomParsers() {
+        XmlTestLoaderConfiguration loaderConfiguration = testClass.getAnnotation(XmlTestLoaderConfiguration.class);
+        if (loaderConfiguration != null) {
+            Arrays.stream(loaderConfiguration.parserConfigurations()).forEach(testCaseParserConfiguration -> {
```
bbortt, line 120: I think streams is superfluous at this point? just use `for(TestCaseParserConfiguration testCaseParserConfiguration: loaderConfiguration.parserConfigurations())`..

core/citrus-spring/src/main/java/com/consol/citrus/config/CitrusNamespaceParserRegistry.java
```diff
@@ -0,0 +1,136 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.xml.BeanDefinitionParser;
+
+/**
+ * Registers bean definition parser for beans in test case, handled by {@link com.consol.citrus.config.handler.CitrusTestcaseNamespaceHandler}
+ *
+ * @author Christoph Deppisch
+ * @since 2007
+ */
+public final class CitrusNamespaceParserRegistry {
```
bbortt, line 34: GitHub does not seem to notice that this has been renamed from `core/citrus-spring/src/main/java/com/consol/citrus/config/TestActionRegistry.java`. have there been any changes or is it just the renaming?

core/citrus-spring/src/main/java/com/consol/citrus/config/handler/CitrusTestcaseNamespaceHandler.java
bbortt: @tschlat so renaming would be fine.. but it's ok if you feel uncomfortable about it.

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
tschlat: As old school programmer it took me a while to adapt to streams but now I am streaming everything ;-).

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
```diff
@@ -127,4 +160,5 @@ public String getSource() {
     public void setSource(String source) {
         this.source = source;
     }
+
```
tschlat, line 163:
```suggestion
```

tschlat:
@bbortt @christophd I have finished my work and the build is green. Time for a last look.

core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java
bbortt: it's not always "the yellow of the egg" 😉

bbortt (approved):

tschlat:
@christophd Magst Du vielleicht noch mal nach dem PR schauen Christoph?

christophd (approved):
Many thanks!

PR #858, opened 2022-05-12
add jdk18 preview build

bbortt:
based on the `jdk17` (current lts) build.


bbortt:
will be included in #904.

PR #875, opened 2022-07-27
fix(#861): Refactor wait action builder

### LINKED ISSUE #861: WaitConditionBuilder constructor never sets the condition in the builder despite taking it as an argument
**Citrus Version**
3.2.1

**Expected behavior**
The constructor should set the condition in the underlying builder.

**Actual behavior**
The constructor does not set the condition in the underlying builder, it has to be set manually.

**Test case sample**
```
Wait.Builder builder = new Wait.Builder()
        .time(timeout);
WaitActionConditionBuilder waitActionConditionBuilder = new WaitActionConditionBuilder(condition, builder);
```
does not work. The ```condition``` will never actually be set. Instead,
```
Wait.Builder builder = new Wait.Builder()
        .time(timeout)
        .condition(condition);

WaitActionConditionBuilder waitActionConditionBuilder = new WaitActionConditionBuilder(condition, waitBuilder);
```
has to be performed. This makes the argument ```condition``` in the ```WaitActionConditionBuilder``` useless.
I'd propose to add a line ```builder.condition(condition)``` in the constructor of ```WaitConditionBuilder```.

christophd:
- Improve constructors to only use builder instead of redundant condition and builder

Fixes #861

### DIFF
--- a/core/citrus-base/src/main/java/com/consol/citrus/container/WaitConditionBuilder.java
+++ b/core/citrus-base/src/main/java/com/consol/citrus/container/WaitConditionBuilder.java
@@ -25,26 +25,20 @@ import com.consol.citrus.condition.Condition;
  * @author Christoph Deppisch
  * @since 2.4
  */
-public abstract class WaitConditionBuilder<T extends Condition, S extends WaitConditionBuilder> implements TestActionBuilder<Wait> {
+public abstract class WaitConditionBuilder<T extends Condition, S extends WaitConditionBuilder<T, S>> implements TestActionBuilder<Wait> {
 
     /** Parent wait action builder */
-    private final Wait.Builder builder;
-
-    /** Condition */
-    private final T condition;
+    private final Wait.Builder<T> builder;
 
     /** Self reference */
     protected S self;
 
     /**
      * Default constructor using fields.
-     * @param condition
      * @param builder
      */
-    public WaitConditionBuilder(T condition, Wait.Builder builder) {
-        this.condition = condition;
+    public WaitConditionBuilder(Wait.Builder<T> builder) {
         this.builder = builder;
-
         this.self = (S) this;
     }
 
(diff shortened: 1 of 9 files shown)


core/citrus-base/src/main/java/com/consol/citrus/container/WaitConditionBuilder.java
```diff
@@ -25,26 +25,20 @@
...
     protected S self;
 
     /**
      * Default constructor using fields.
-     * @param condition
      * @param builder
      */
-    public WaitConditionBuilder(T condition, Wait.Builder builder) {
-        this.condition = condition;
+    public WaitConditionBuilder(Wait.Builder<T> builder) {
```
bbortt, line 40: the problem is, this change is potentially breaking @christophd.. should we target it for citrus 4.0? i am working on that, but have a silly error with `citrus-docker` and don't know what's up there.

bbortt: yes, it is never backwards compatible as soon as you just remove an argument from a constructor. if someone uses this constructor, he's doomed :(

bbortt: @christophd actually, I was just able to resolve the `citrus-docker` compilation failure.. maybe I can manage to provide a fullbuild of citrus@v4.0.0 until this evening.

christophd: @bbortt to be honest I think it would be ok to change the constructor even on a minor version upgrade. I know people may need to adjust their code but the change is trivial and the API change is done on a kind of internal API. People using the end user Java DSL will not even notice when updating. 

I know most of the libraries are doing it this way when the changes are trivial and when there is a notice of the change added to the release notes (that we do not have at the moment which is a problem).

Summing up I would add this for upcoming version 3.3.0 if we agree to have a decent release notes section.

bbortt: I am fine with that. did also note it in #878.

bbortt (approved):

PR #886, opened 2022-09-06
default test loader supports fluent configuration

### LINKED ISSUE #885: new `DefaultTestLoader` does not support default arguments
@christophd the new [`DefaultTestLoader`](https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/com/consol/citrus/common/DefaultTestLoader.java) has just an empty constructor. the old one (`XmlTestLoader`) [had some arguments to it](https://github.com/citrusframework/citrus/blob/v3.2.1/core/citrus-spring/src/main/java/com/consol/citrus/common/XmlTestLoader.java#L59).

should I add a second constructor or would you prefer a builder pattern?

e.g.

```java
    public DefaultTestLoader() {
        // Standard constructor
    }

    public DefaultTestLoader(Class<?> testClass, String testName, String packageName, CitrusContext citrusContext) {
        this.testClass = testClass;
        this.testName = testName;
        this.packageName = packageName;
        this.citrusContext = citrusContext;
    }
```

or

```java
    public DefaultTestLoader [with]TestClass(Class<?> testClass) {
        this.testClass = testClass;
        return this;
    }

    // etc...
```

I am unsure, which one does look nicer? or should I add both?

bbortt:
closes #885. see issue description for details.

- [ ] await decision


bbortt:
oh, and side note for myself: I think we should have builder methods foreach method. or constructors foreach "combination" of parameters. I'll work on this.

bbortt:
@christophd I've signed the commits, this PR is ready for review. would be cool to have it release as `3.3.1` afterwards 🚀

core/citrus-spring/src/main/java/com/consol/citrus/common/SpringXmlTestLoader.java
```diff
@@ -43,6 +43,16 @@ public class SpringXmlTestLoader extends DefaultTestLoader implements TestSource
 
     private String source;
 
+    @Override
+    protected void initializeTestRunner() {
+        if (citrusContext == null) {
+            ApplicationContext applicationContext = loadApplicationContext();
+            citrusContext = CitrusSpringContext.create(applicationContext);
```
christophd, line 50: Not sure about this one. The `CitrusContext` should have been injected via `@CitrusResource` annotation at this point. If this is not the case for any reason the `CitrusContext` should be retrieved from `Citrus` instance (also being injected) rather than creating it via `loadApplicationContext()`.

If `CitrusContext` is not set at this point `loadApplicationContext()` will not be able to retrieve the parent application context properly and create an empty application context. I do not think that this is what we want.

So if `CitrusContext` is not set at this point for some reason I would try to retrieve it from `Citrus` instance as it is done in `super.initializeTestRunner()`. Otherwise an error should be thrown as it is already done in `super.initializeTestRunner()`.

WDYT?

core/citrus-base/src/main/java/com/consol/citrus/common/DefaultTestLoader.java
```diff
@@ -108,16 +125,22 @@ protected void doLoad() {
     protected void initializeTestRunner() {
         if (runner == null) {
             if (context == null) {
-                if (citrus == null) {
-                    throw new CitrusRuntimeException("Missing Citrus framework instance for loading test");
+                if (citrusContext == null) {
```
christophd, line 128: Nice finding! thx!

christophd (commented):
Many thanks! LGTM except that I would throw exception rather than creating a new `CitrusContext` (with empty Application context) in SpringXmlTestLoader

core/citrus-spring/src/main/java/com/consol/citrus/common/SpringXmlTestLoader.java
bbortt: oh, I see. if `citrusContext` is null at this point, then `applicationContext` will also be `null`. I'll revert this change, then try again to implement the next citrus release into our applications. if it doesn't work, I'll submit another PR.

christophd (approved):

PR #888, opened 2022-09-09
provide standalone bom for citrus

### LINKED ISSUE #887: citrus-bom is not a standalone BOM
The maven artifact com.consol.citrus:citrus-bom is a child-pom of citrus-catalog and the citrus "root"-pom. Since the com.consol.citrus:citrus pom manages all dependencies for citrus, they are inherited to the citrus-bom artefact. As a result, importing the citrus-bom imports a bunch of other dependencies into a project:

```
			<dependency>
				<groupId>com.consol.citrus</groupId>
				<artifactId>citrus-bom</artifactId>
				<version>3.3.0</version>
				<scope>import</scope>
				<type>pom</type>
			</dependency>
```

Have a look at the "Managed Dependencies" section in https://mvnrepository.com/artifact/com.consol.citrus/citrus-bom/3.3.0, which manages versions for spring, jetty, a bunch of apache projects and many other projects. When importing the citrus-bom, I only want it to manage citrus artifacts, because otherwise, updating citrus could have unwanted side effects for my project, which are not obvious.

As a solution, the citrus-bom artifact should be built outside the maven multimodule-hierarchy of citrus. As a reference, have a look at the junit-bom: https://mvnrepository.com/artifact/org.junit/junit-bom/5.9.0

boskoop:
closes #887, see description for details

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,3 +61,4 @@ jobs:
       - name: Build Citrus
         run: |
           ./mvnw --no-transfer-progress install
+          ./mvnw --no-transfer-progress -f bom/pom.xml install
--- a/.github/workflows/lts.yml
+++ b/.github/workflows/lts.yml
@@ -62,3 +62,4 @@ jobs:
       - name: Build Citrus
         run: |
           ./mvnw --no-transfer-progress -Djava.version=17 install
+          ./mvnw --no-transfer-progress -Djava.version=17 -f bom/pom.xml install
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,18 +1,70 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-
-  <parent>
-    <artifactId>citrus-catalog</artifactId>
-    <groupId>com.consol.citrus</groupId>
-    <version>3.4.0-SNAPSHOT</version>
-    <relativePath>../pom.xml</relativePath>
-  </parent>
-
+  <groupId>com.consol.citrus</groupId>
   <artifactId>citrus-bom</artifactId>
+  <version>3.4.0-SNAPSHOT</version>
   <packaging>pom</packaging>
-  <name>Citrus :: Catalog :: BOM</name>
-
+  <name>Citrus (Bill of Materials)</name>
+  <description>This Bill of Materials POM can be used to ease dependency management when referencing multiple Citrus artifacts using Gradle or Maven.</description>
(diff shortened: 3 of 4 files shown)


bbortt (approved):
I am not entirely sure if this will still work with the release scripts. please ask @christophd for his opinion.

/lgtm

boskoop:
> I am not entirely sure if this will still work with the release scripts. please ask @christophd for his opinion.
> 
> /lgtm

Yeah, I wasn't sure either. I think it should work since the release scripts traverse all directories for updating and dont't use e.g. the maven versions plugin. But I'm not sure about how publishing will work.

catalog/pom.xml
```diff
@@ -17,7 +17,6 @@
   <modules>
     <module>citrus-core</module>
     <module>citrus-endpoint-catalog</module>
-    <module>citrus-bom</module>
```
christophd, line 20: we can still keep this as a submodule also when there is no parent relationship anymore. this way we make sure the BOM is built with the project and that the BOM gets published to Maven central during the release process.

.github/workflows/lts.yml
```diff
@@ -62,3 +62,4 @@ jobs:
       - name: Build Citrus
         run: |
           ./mvnw --no-transfer-progress -Djava.version=17 install
+          ./mvnw --no-transfer-progress -Djava.version=17 -f bom/pom.xml install
```
christophd, line 65: no need for this if we keep the BOM as a submodule (see comment below)

.github/workflows/build.yml
```diff
@@ -61,3 +61,4 @@ jobs:
       - name: Build Citrus
         run: |
           ./mvnw --no-transfer-progress install
+          ./mvnw --no-transfer-progress -f bom/pom.xml install
```
christophd, line 64: no need for this if we keep the BOM as a submodule (see comment below)

bom/pom.xml
```diff
@@ -1,18 +1,70 @@
...
+    <license>
+      <name>The Apache License, Version 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <distribution>manual</distribution>
+      <comments>Citrus is published free of charge under Apache License, Version 2.0. You are allowed to use Citrus to test any kind of software
+      (also commercial applications). If you do this, we kindly ask you to tell us about it so that we can add you to our references
+      list. You may also use our "Tested using Citrus" logo in your online or printed documentation.</comments>
+    </license>
+  </licenses>
+  <developers>
```
christophd, line 21: I think we can skip the developers section here

christophd (commented):
Many thanks! looks good except that I would like to keep it as a submodule so that the BOM is built and released as part of the project.

Also as a long term solution I would like to separate the citrus parent POM into two POMs
- citrus POM just holding the common information such as license, url, developers, scm
- citrus-parent POM being a child of citrus POM adding the dependency management

This way we can use either citrus or citrus-parent as a parent POM in submodules

christophd:
@boskoop are you able to further work on this issue in order to rebase and adjust some changes as requested? Many thanks!

Issue #846, opened 2022-03-12
Response not logged in case of error in extract


chokdee:
**Citrus Version**
3.2.0

**Expected behavior**
When sending a Request and call the receive() the response should be logged in any case

**Actual behavior**
If the extract failed because the response didn't have the expected body, the Response isn't logged. 
That make the analysis very difficult.

**Test case sample**
```
 $(http()
        .client(myClient)
        .send()
        .post("/something")
        .message()
        .header("Accept", "*/*")
        .header("Authorization", "Bearer ${access_token}")
        .header("Content-Type", "application/json")
        .body(new ClassPathResource("test.json")));
$(http()
        .client(myClient)
        .receive()
        .response(HttpStatus.OK)
        .message()
        .type(MessageType.JSON)
        .extract(JsonPathSupport.jsonPath().expression("value", "${myvalue}")));
```

christophd:
+1

PR #891, opened 2022-09-27
V3.4 bugfix for 846

chokdee:
This is a fix for the following issue
https://github.com/citrusframework/citrus/issues/846


bbortt (approved):
/lgtm

christophd:
@chokdee Thank you for the PR!

I think your fix resolves the situation only for Http based communication. IMO the problem is more in the receive action in general. The message content gets logged in the message validator implementations before validation takes place. For example here: 

https://github.com/citrusframework/citrus/blob/main/validation/citrus-validation-text/src/main/java/com/consol/citrus/validation/text/PlainTextMessageValidator.java#L61

But things like variable extraction takes place before that message validation in the receive action. So the proper fix would be to log the received message content right after it has been received in received action like here:

https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/com/consol/citrus/actions/ReceiveMessageAction.java#L167

I think I also recall why the message content logging has been moved to the message validator implementations. This is because the message content may be `null` or it may be non textual content such as binary content. The message validator implementation may know the content type and skip logging when appropriate.

In case we log the content in receive action I think we need to add some check that the message content is not null and that it is printable content before logging. WDYT?

Also we should make sure to not log each message multiple times (e.g. in Http logging interceptor, receive action and message validator)

christophd (commented):
Many thanks! I have added some comment. see below

PR #895, opened 2022-09-28
fix(#894): Avoid NullpointerException in citrus:env function

### LINKED ISSUE #894: citrus:env function causes Nullpointer
**Citrus Version**
3.3.x

Using the EnvironmentPropertyFunction leads to NullpointerException because Spring environment is not set properly. The function is loaded via resource path lookup and this is missing the injection of the Spring environment instance leading to NPEs

christophd:
Make sure to properly set Spring environment and application context on all functions in default function library. Adding factory bean for default function library that is aware of application context and Spring environment for proper dependency injection. Also, automatically add all functions to the function library that live as Spring beans in the application context.

Use traditional system environment variable settings lookup as fallback when Spring environment is not set properly.

Fixes #894

### DIFF
--- a/core/citrus-spring/src/main/java/com/consol/citrus/functions/DefaultFunctionLibraryFactory.java
+++ b/core/citrus-spring/src/main/java/com/consol/citrus/functions/DefaultFunctionLibraryFactory.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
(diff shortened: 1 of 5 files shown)


bbortt (approved):
that is indeed very useful. actually stumbled upon this a while ago and added an fix in our company-internal branch. never made it to the open source project, because I forgot xD but /lgtm!

Issue #900, opened 2022-10-24
Add OWASP dependency check


ioggstream:
**User story**
As a [developer], I want to [check the security of dependencies] so that [the software is more secure]

**Additional context**

Run org.owasp:dependency-check-maven:check

ioggstream:
See https://github.com/ioggstream/citrus/actions/runs/3313241276/jobs/5470950802#step:7:5995

PR #901, opened 2022-10-24
Fix: #900. Add dependency checks

ioggstream:
## This PR

- adds informative owasp dependency checks in CI

## Note

you can enforce validation output and block deployment when insecure deps are found

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,3 +61,6 @@ jobs:
       - name: Build Citrus
         run: |
           ./mvnw --no-transfer-progress install
+      - name: Informative security dependency checks.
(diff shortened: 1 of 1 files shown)


bbortt (approved):
I like this. I still think we should consider adding @dependabot @christophd.. don't remember what the suspicions were :D

christophd:
+1 but I would rather run this in a separate CI job. making this check part of the general build job may break PRs from people that are not in charge of maintaining the dependencies.

Also what about running this on a nightly basis instead of with each PR or code change?

christophd:
uff, this PR is open for a long time with no activity. Closing ...

Issue #934, opened 2023-03-22
Citrus w/ jUnit5: Neither an actual suite name nor groups are passed to BeforeSuites / AfterSuites shouldExecute method


tko-qs:
**Citrus Version**
latest (main-Branch) / 3.x with jUnit5

**User story / Expected behavior** 
If one wants to use `beforeSuite` (`com.consol.citrus.container.BeforeSuite`) and `afterSuite` (`com.consol.citrus.container.AfterSuite`) to prepare the execution of tests, e.g. purge some queues or prepare SUT by applying ITest configuration, one also wants to be able to be able to use the selective execution behavior by implementing the method `shouldExecute(String suiteName, String[] includedGroups)` that both interfaces provide - regadlessly of the runtime executing the tests.

**Actual current behavior**
With jUnit5 the CitrusExtension does not pass an actual suite name to the method `shouldExecute` and neither passes the tags (equivalent to groups in jUnit5).

See:
https://github.com/citrusframework/citrus/blob/52c1daa3b725d219c744471741fdd70d052ae105/core/citrus-api/src/main/java/com/consol/citrus/container/BeforeSuite.java#L14
https://github.com/citrusframework/citrus/blob/52c1daa3b725d219c744471741fdd70d052ae105/core/citrus-api/src/main/java/com/consol/citrus/container/AfterSuite.java#L14
https://github.com/citrusframework/citrus/blob/52c1daa3b725d219c744471741fdd70d052ae105/runtime/citrus-junit5/src/main/java/com/consol/citrus/junit/jupiter/CitrusExtension.java#L79
https://github.com/citrusframework/citrus/blob/52c1daa3b725d219c744471741fdd70d052ae105/runtime/citrus-junit5/src/main/java/com/consol/citrus/junit/jupiter/CitrusExtension.java#L87

**Workaround**
AFAIK there is no actual workaround to passing those values to the method. In the current context I am using Citurs the groups that shall be executed are injected into the `mvn` command anyway, so I can check the property via `System.getProperty` in `shouldExecute`, but it is just applicable for "groups", not a test class / suite.

**Suggested solution**
Pass some values to the method when delegating to the individual callbacks in `com.consol.citrus.junit.jupiter.CitrusExtension#beforeAll` and `com.consol.citrus.junit.jupiter.CitrusExtension#afterAll`.
I will open a pull request with the suggested changes. It currently looks like all tests are still passing after implementing it.

PR #935, opened 2023-03-22
Citrus w/ jUnit5: Neither an actual suite name nor groups are passed to BeforeSuites / AfterSuites shouldExecute method

tko-qs:
Changed the suite name passed to beforeSuite and afterSuite callbacks to the actual class or method name and also passed the tags of the current test case or method to the call. The default Citrus suite name that comes from a constant will still be used as fallback.

Related issue: #934


christophd:
many thanks!

christophd:
Fixes #934

Issue #922, opened 2023-01-31
Citrus v4.0.0


christophd:
We should start thinking and discussing about a new major version of Citrus 4.0.0.

I suppose to migrate main branch to 4.0.0-SNAPSHOT and start working on that major version soon. At the same time update branch 3.x and continue with Citrus v3 there

Checklist for Citrus 4.0.0:

- [x] Use 4.0.0-SNAPSHOT on main branch
- [x] Use Java 17 as minimum requirement
- [x] Remove deprecated code
- [x] Remove vintage modules (e.g. citrus-arquillian, citrus-java-dsl)
- [x] Update to Spring 6.0 (see #907)
- [x] Update to Camel 4.0
- [x] Migrate from `javax.*` to `jakarta.*` (JEE10)
- [x] Move groupId from `com.consol.citrus` to `org.citrusframework`
- [ ] Update/Rewrite documentation
- [x] Update Jetty to v11 (see #871)
- [x] Move citrus-remote modules to separate repository
- [x] Move citrus-jdbc module to [citrus-db](https://github.com/citrusframework/citrus-db) repository

Optional:

- [ ] Remove support for JUnit4
- [ ] Update citrus-kubernetes module to use latest Kubernetes client
- [ ] Update citrus-docker module
- [ ] Remove citrus-restdocs module
- [ ] Move from TestNG to JUnit Jupiter (internal refactoring)

christophd:
Things to discuss:

- Remove support for JUnit4
- Discontinue citrus-restdocs module
- Discontinue citrus-remote module
- Discontinue citrus-jmx module
- Discontinue citrus-rmi module
- Update citrus-kubernetes module to use latest Kubernetes client
- Update citrus-docker module

bbortt:
I'm in 😉 #907 will break things pretty hard. and I can certainly work on this stuff, but only part-time. so, any help is much appreciated.

bbortt:
also #871.

turing85:
> Things to discuss:
> 
> ...
> * Discontinue citrus-remote module
> ...
> * Update citrus-docker module

What is the rationale for this, and what would be the alternative (especially when using YAKS is not an option?

bbortt:
#904 will solve the following:

> - [ ] Use Java 17 as minimum requirement
> - [ ] Update to Spring 6.0 (see https://github.com/citrusframework/citrus/issues/907)

christophd:
[...]

PR #936, opened 2023-03-23
fix(#922): Move citrus-jdbc module to citrus-db project

christophd:
Moved module to [citrus-db](https://github.com/citrusframework/citrus-db) (see https://github.com/citrusframework/citrus-db/pull/58)


bbortt:
cool! 🚀

PR #946, opened 2023-04-24
remove vintage modules

bbortt:
discontinue support of the `citrus-arquillian` and
`citrus-java-dsl` module.


pom.xml
```diff
@@ -515,11 +514,6 @@
         <artifactId>junit-jupiter-engine</artifactId>
         <version>${junit.jupiter.version}</version>
       </dependency>
-      <dependency>
-        <groupId>org.junit.vintage</groupId>
-        <artifactId>junit-vintage-engine</artifactId>
-        <version>${junit.jupiter.version}</version>
-      </dependency>
```
bbortt, line 522: apparently I have deleted a bit too much xD that is why the build fails. anyway, this should be gone too.. but we've already mentioned this within the junit migration, I think.

PR #947, opened 2023-05-17
Detect if the current execution is started from a jar, and if so analyze the jar for resources

### LINKED ISSUE #862: default validators not present in remote-jar
Citrus Version : 3.2.1

---
Expected behavior:

When we add a validator to our project (e.g. `citrus-validation-text`), a default validator should be provided and no further configuration is needed.

---
Actual behavior:

Tests fail with the message 

    Unable to find proper message validator. Message type is 'PLAINTEXT' and message payload is ...

---
Reproducer:

- checkout [this git-repository](https://github.com/turing85/citrus-playground), switch to branch `default-validator-not-working`
- build service and tests: 

      ./mvnw -Pbuild-citrus-jar package
- deploy the service: 

      cd local-deployment && docker compose up -d service && cd ..
- trigger the tests through failsafe: 

      ./mvnw -pl citrus verify
- observe that the tests succeed
- deploy the citrus tests: 

      cd local-deployment && docker compose up -d citrus-tests && cd ..
- trigger the tests through the `citrus-maven-remote-plugin`: 

      ./mvn -Ptrigger-citrus-remote -pl citrus verify
- observe that the tests fail

---
Logs of the two runs on my local machine:

[citrus-reports-from-remote-failing.tar.gz](https://github.com/citrusframework/citrus/files/8818352/citrus-reports-from-remote-failing.tar.gz)
[citrus-reports-from-surefire-successful.tar.gz](https://github.com/citrusframework/citrus/files/8818353/citrus-reports-from-surefire-successful.tar.gz)

---
Additional information:

1. This only happens when a remote jar (and possibly war,  haven't tested this) is deployed. Local execution through failsafe works as expected.
1. We can make the tests succeed by un-commenting [this part right here](https://github.com/turing85/citrus-playground/blob/default-validators-not-working/citrus/src/test/java/de/turing85/citrus/tests/citrus/configuration/ConfigurationRoot.java#L7) (this will lead to [this file](https://github.com/turing85/citrus-playground/blob/default-validators-not-working/citrus/src/test/java/de/turing85/citrus/tests/citrus/configuration/Validators.java) being included, which defines a default validator. 
1. In the server logs of the remote execution, I see

       2022-06-01 19:38:52.123  INFO --- [pool-2-thread-1] trationDelegate$BeanPostProcessorChecker : Bean 'referenceResolver' of type [com.consol.citrus.context.SpringBeanReferenceResolver] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
    Maybe this is related.

turing85:
Fixes #862

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/spi/ResourcePathTypeResolver.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/spi/ResourcePathTypeResolver.java
@@ -1,20 +1,29 @@
 package org.citrusframework.spi;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
+import java.net.URL;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
 
 import org.citrusframework.exceptions.CitrusRuntimeException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.UrlResource;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.core.io.support.ResourcePatternResolver;
 
@@ -40,6 +49,9 @@ public class ResourcePathTypeResolver implements TypeResolver {
 
     /** Base path for resources */
     private final String resourceBasePath;
+
+    /** Zip entries as String, so the archive is read only once */
+    private final List<String> zipEntriesAsString = new ArrayList<>();
 
     /**
      * Default constructor using META-INF resource base path.
@@ -99,8 +111,10 @@ public class ResourcePathTypeResolver implements TypeResolver {
         final String path = getFullResourcePath(resourcePath);
 
         try {
-            Stream.of(new PathMatchingResourcePatternResolver().getResources(ResourcePatternResolver.CLASSPATH_ALL_URL_PREFIX + path + "/*"))
-                    .forEach(file -> {
+            Stream.concat(
+                    Stream.of(new PathMatchingResourcePatternResolver().getResources(ResourcePatternResolver.CLASSPATH_ALL_URL_PREFIX + path + "/*")),
+                    resolveAllFromJar(path))
+                .forEach(file -> {
                         Optional<String> resourceName = Optional.ofNullable(file.getFilename());
                         if (resourceName.isEmpty()) {
                             LOG.warn(String.format("Skip unsupported resource '%s' for resource lookup", file));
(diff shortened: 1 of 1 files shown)


core/citrus-api/src/main/java/com/consol/citrus/spi/ResourcePathTypeResolver.java
```diff
@@ -130,6 +141,32 @@ public <T> Map<String, T> resolveAll(String resourcePath, String property, Strin
         return resources;
     }
 
+    private static List<Resource> resolveAllFromJar(String path) {
+        URL root = ResourcePathTypeResolver.class.getProtectionDomain().getCodeSource().getLocation();
+        List<Resource> result = new ArrayList<>();
+        String rootAsString = root.toString();
+        if (rootAsString.endsWith(".jar") && !rootAsString.matches(".*" + File.separator + "citrus-api-\\d+\\.\\d+\\.\\d+(-.*)?\\.jar")) {
+            try (ZipInputStream in = new ZipInputStream(root.openStream())) {
+                while (true) {
```
bbortt, line 150: thanks very much for the contribution (and the detailed description in the issue)! there is just one thing I would like to discuss. the `while(true)` part looks ugly in my opinion. what about this?

```java
ZipEntry entry;
while ((entry = in.getNextEntry()) != null) { // or !Objects#isNull if you prefer that
    if (entry.getName().startsWith(path)) {
        Optional.of(ResourcePathTypeResolver.class)
            .map(Class::getClassLoader)
            .map(classLoader -> classLoader.getResource(entry.getName()))
            .map(UrlResource::new)
            .ifPresent(result::add);
    }
}
```

turing85: This won't work since the usage of `entry` in the lambda needs it to be final or effectively-final. We would need to rewrite the `Optional`-part.

bbortt: that is true, I forgot that. you could use an [atomic reference](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReference.html) for this purpose, which you can `set` and `get`. but that again might be too much trouble for what it is 🙁 thoughts @christophd?

turing85: @bbortt I refactored the code a little (added caching for the zip entry so the zip is not read multiple times). This moved the problematic code in a non-lambda part and I applied your change. Care to take another look?

core/citrus-api/src/main/java/com/consol/citrus/spi/ResourcePathTypeResolver.java
```diff
@@ -130,6 +143,37 @@ public <T> Map<String, T> resolveAll(String resourcePath, String property, Strin
         return resources;
     }
 
+    private List<Resource> resolveAllFromJar(String path) {
+        List<Resource> result = new ArrayList<>();
```
bbortt, line 147: if you would return the stream starting at `getZipEntries().stream()` and then use lambdas to do the mapping and filtering accordingly you could event skip this resource allocation. did you think about that?

turing85: I am not sure what exactly you mean. Do you mean returning a `Stream<Resource>` instead of a `List<Resource>`?

turing85: @bbortt okay I think I get what you meant. Updated the PR.

bbortt: that is exactly what I meant! 😉 could you test this with a local installation?

turing85: Already did. Works fine.

bbortt:
ok, now the only thing is I don't know about the backport to [citrusframework:3.x](https://github.com/citrusframework/citrus/tree/3.x). ping @christophd?

turing85:
> ok, now the only thing is I don't know about the backport to [citrusframework:3.x](https://github.com/citrusframework/citrus/tree/3.x). ping @christophd?

@bbortt As far as I know, this fix is only relevant for `citrus-remote`. And as far as I know, `citrus-remote` is discontinued in `citrus` version `4.x`. 

I can rebase onto the current `main`, but a backport to `3.x` would be nice, as well as a new `3.x` release. I can take care of the backport. All test succeed on `main` as well. I cannot test this locally however for citrus `4.x` however.

I'll rebase if you give me the go-ahead and when there are no further things to change.

bbortt:
> as far as I know, citrus-remote is discontinued in citrus version 4.x.

that is true, deleted this module already. still, the core remains active in the 4x release, of course.

> I cannot test this locally for citrus 4.x.

no worries, we'll just keep this in mind and come back to it before going forward with milestones.

turing85:
@bbortt rebased on `main`, changed base for PR to `main`. I'll open the downstream PR as soon as this PR is merged.

turing85:
So... are we gonna merge this PR or nah?

bbortt:
> So... are we gonna merge this PR or nah?

sorry, lots to do! thanks for the reminder and your patience ❤️

bbortt (approved):

bbortt:
@christophd why does the workflow not run? O.o

christophd:
@bbortt @turing85 sorry, been on holidays. having a look this week

christophd:
no idea why the workflow has not been kicked off. I just merged now let's see how it goes ...

thanks @turing85 and @bbortt for your patience

bbortt:
thank you @christophd, hope you enjoyed the days off 😉

PR #954, opened 2023-07-07
chore(#922): Add unit test coverage from removed java-dsl module

christophd:
- Bring back some unit test coverage from removed java-dsl module
- Add TestBehaviorExecutingEndpointAdapter as a successor of former Java DSL test runner executing endpoint adapter

### DIFF
--- a/connectors/citrus-docker/src/test/java/org/citrusframework/docker/actions/dsl/DockerTestActionBuilderTest.java
+++ b/connectors/citrus-docker/src/test/java/org/citrusframework/docker/actions/dsl/DockerTestActionBuilderTest.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2006-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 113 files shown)


bbortt (approved):
cool stuff!

connectors/citrus-docker/src/test/java/org/citrusframework/docker/actions/dsl/DockerTestActionBuilderTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.citrusframework.docker.actions.DockerExecuteAction.Builder.docker;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Christoph Deppisch
+ * @since 2.4
+ */
+public class DockerTestActionBuilderTest extends UnitTestSupport {
```
bbortt, line 48: I think test classes, as well as methods, should not be declared public.

christophd: Yeah, thanks. Probably something to add to the unit test refactoring task that targets to use JUnit Jupiter instead of TestNG

PR #959, opened 2023-07-29
fullbuild on windows

bbortt:
citrus sticks to `\n` for line breaks (linux style), that did not work when building on windows (`\r\n`): some tests using `System.lineSeparator()` or `System.getProperty("line.separator")` did obviously fail.

I initially thought to search and replace the `\n` in the source, but then changed direction, because I find it consistent if citrus **always** uses `\n` for linebreaks, disregarding the underlying system.


bbortt:
@christophd someone now the build broke.. why didn't it run on top of #961? probably some rules. will take a look at it this evening. this PR is a must-have for M2.

christophd:
@bbortt yes, confusing. No hurry, I will wait with M2 release until this is solved and merged

christophd (approved):
Looks good. thanks a lot

bbortt:
@christophd It's working now. It looks like I've missed running the tests while looking only at the simulator 😅 turns out, some tests regarding `citrus-validation-json` were no longer compiling at all. I did some lookup and saw that the used implementation is quite old: I've replace it in 39c16b70ebd814f17bf07fabca4e82cc73595d90. that commit went in between the existing two. afterwards, 4c0833ffd3abd184d91407448a29f3034bcfb7c3 (the existing "compatibility fixes") went through without any problems.

in addition to that and because of our above discussion, I've added 8f6822e32c7830f3076e460be8b58fc9d882464a in the end.

christophd:
@bbortt it looks good! I have pushed another change so we use consistent eclipse angus-activation library in all dependencies. 

Also I brought back jakarta-activation API and implementation version in dependency management as it is being used by many transitive dependencies. This makes sure that we at least have the same version of activation implementations on classpath.

I think we are not ready to merge it

BTW have I already mentioned that `javax.*` vs. `jakarta.*` is a complete mess. Gosh darn it!

bbortt:
> BTW have I already mentioned that javax.* vs. jakarta.* is a complete mess. Gosh darn it!

It's for sure! Completely fucks me up every time.

christophd:
@bbortt do you want to check one more time on the citrus-simulator project before doing the M2 release?

bbortt:
> @bbortt do you want to check one more time on the citrus-simulator project before doing the M2 release?

sure. but until next Monday my schedule is pretty tight... if that's ok for you?

PR #970, opened 2023-08-21
Remove Travis reference from Contributing guide

apupier:


bbortt (approved):
that makes sense lmao

PR #972, opened 2023-09-05
feat: static access to extractor builders

bbortt:
you already have access to `fromHeaders` and `fromBody`, but not for JSON or XPaths.


bbortt:
> so the Java DSL usage would be receive().extract(fromJsonPath().expression(...))

correct.

> Just FYI there is also org.citrusframework.dsl.JsonPathSupport.jsonPath() which is able to act as an extractor in receive().extract(jsonPath().expression(...))

didn't know that. still, after looking [at the source](https://github.com/citrusframework/citrus/blob/3cf248087588acd84f50c5694e193c1600c57770/validation/citrus-validation-json/src/main/java/org/citrusframework/dsl/JsonPathSupport.java#L57) it would be `jsonPath().asExtractor().expression(...)` - a bit too long for my gusto 😉

christophd:
The `asExtractor()` is called automatically when the expression is being used as an extractor, so you can just use `jsonPath().expression(...)` 😄

So you can use the same static `jsonPath()` builder method and use it as extractor, validator, processor, ... which is good because you only have to import one single static builder method for all jsonPath features.

Issue #973, opened 2023-09-07
Allow standard Reporter overrides


tschlat:
**User story**
As an active user and developer within the Citrus framework ecosystem, where I contribute additional features built upon the foundation of Citrus, I am seeking the capability to customize and override Citrus default logging mechanisms. For example I want to be able to provide a more sophisticated logging via a specific LoggingReporter. I am explicitly seeking for a solution that allows me to **replace** default logging behaviour, so that I do not get things reported twice. In addition, I want to avoid bean overriding in Spring.

**Additional context**
At present, Spring's standard reporters are managed by the ReporterConfig class. Overriding these reporter beans typically involves setting spring.main.allow-bean-definition-overriding=true plus using the @Primary annotation on the replacement bean. However, this approach is discouraged by Spring due to potential side effects. To address this issue more elegantly, there is a need for an enhanced solution that facilitates overriding while minimizing any undesirable consequences.

The standard reports in focus are:
**LoggingReporter**
**HtmlReporter**
**JUnitReporter**

The LoggingReporter serves a multifaceted purpose beyond the scope of mere test result reporting. Consequently, removing it entirely is not a viable option. In contrast, both the HtmlReporter and JUnitReporter already offer a mechanism for disabling their functionality. Hence, I propose that introducing a disablement option for the standard LoggingReporter would effectively meet the requirements. This would allow users to replace standard reporters with custom logic or other reporting solutions using a simple configuration snippet like the following:

```java
@Bean
public CustomLoggingReporter customLoggingReporter(@Qualifier("citrusLoggingReporter") LoggingReporter citrusLoggingReporter) {
    citrusLoggingReporter.setEnabled(false);
    return new CustomLoggingReporter();
}
```

PR #974, opened 2023-09-07
feat(#973): Allow standard reporter override

tschlat:
I would like to provide a solution for the respective issue.

Note: I am not sure about the comment, I added to the docs as I am not familliar with the configuration of Citrus context in the case that Spring is not used.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/report/DefaultTestReporters.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/report/DefaultTestReporters.java
@@ -1,7 +1,6 @@
 package org.citrusframework.report;
 
-import java.util.Arrays;
-import java.util.List;
+import java.util.Map;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


core/citrus-base/src/main/java/org/citrusframework/report/DefaultTestReporters.java
```diff
@@ -1,21 +1,18 @@
...
 public class DefaultTestReporters extends TestReporters {
 
-    public static final List<TestReporter> DEFAULT_REPORTERS = Arrays.asList(
-            new LoggingReporter(),
-            new HtmlReporter(),
-            new JUnitReporter()
-    );
+    public static final Map<String, TestReporter> DEFAULT_REPORTERS = Map.of("citrusLoggingReporter", new LoggingReporter(),
+        "citrusHtmlReporter",new HtmlReporter(),
+        "citrusJunitReporter",new JUnitReporter());
```
bbortt, line 13:
```suggestion
    public static final Map<String, TestReporter> DEFAULT_REPORTERS = Map.of(
        "citrusLoggingReporter", new LoggingReporter(),
        "citrusHtmlReporter", new HtmlReporter(),
        "citrusJunitReporter", new JUnitReporter()
    );
```

bbortt:
I think you'll also need to fix some tests: https://github.com/citrusframework/citrus/actions/runs/6106192495/job/16570862788?pr=974#step:6:1751.

tschlat:
> I think you'll also need to fix some tests: https://github.com/citrusframework/citrus/actions/runs/6106192495/job/16570862788?pr=974#step:6:1751.

tschlat:
> > I think you'll also need to fix some tests: https://github.com/citrusframework/citrus/actions/runs/6106192495/job/16570862788?pr=974#step:6:1751.

I saw this comming. Now the beans are missing and cannot be autowired. Unforunately, these Reporters also implement some other interfaces like TestListener, TestSuiteListener, TestActionListener and MessageListener.  
So I see, that this approach is suboptimal.

bbortt:
closes https://github.com/citrusframework/citrus/issues/973.

tschlat:
Closed in favour of: [973](https://github.com/citrusframework/citrus/pull/973)

PR #975, opened 2023-09-07
feat(#973): Allow deactivation of standard LoggingReporter

tschlat:
I came up with a solution that enables the complete disablement of the LoggingReporter's functionality. Considering that both the HtmlReporter and JUnitReporter already offer similar disablement options, I believe this solution adequately addresses the requirements.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/report/LoggingReporter.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/report/LoggingReporter.java
@@ -27,23 +27,39 @@ import org.citrusframework.context.TestContext;
 import org.citrusframework.message.Message;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.helpers.NOPLoggerFactory;
 import org.springframework.util.StringUtils;
 
 /**
  * Simple logging reporter printing test start and ending to the console/logger.
+ * <p/>
+ * Implementation note: The disablement of the reporter is achieved by using a {@link org.slf4j.helpers.NOPLogger},
+ * meaning that this class should primarily focus on logging operations and not extend beyond that functionality.
  *
  * @author Christoph Deppisch
  */
 public class LoggingReporter extends AbstractTestReporter implements MessageListener, TestSuiteListener, TestListener, TestActionListener {
 
     /** Inbound message logger */
-    private static final Logger INBOUND_MSG_LOGGER = LoggerFactory.getLogger("Logger.Message_IN");
+    private static Logger inboundMessageLogger = LoggerFactory.getLogger("Logger.Message_IN");
+
+    /** The inbound message logger used when the reporter is enabled */
+    private static final Logger enabledInboundMessageLogger = inboundMessageLogger;
 
     /** Outbound message logger */
-    private static final Logger OUTBOUND_MSG_LOGGER = LoggerFactory.getLogger("Logger.Message_OUT");
+    private static Logger outboundMessageLogger = LoggerFactory.getLogger("Logger.Message_OUT");
+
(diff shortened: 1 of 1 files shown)


core/citrus-base/src/main/java/org/citrusframework/report/LoggingReporter.java
```diff
@@ -282,6 +329,18 @@ protected void debug(String line) {
...
-        return log.isDebugEnabled();
+        return enabled && log.isDebugEnabled();
+    }
+
+    /**
+     * Sets the enabled property.
+     * @param enabled
+     */
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
```
bbortt, line 340: is there a possibility to completly supress the underlying logger? because you'd have to add an immediate `return` to **every** method - also `info` and `error` etc.

bbortt: or we do say that in the case of a direct invocation it might be wanted and we do knowingly not suppress it. but that clashes with the current implementation of the `debug` method.

bbortt: maybe this https://www.slf4j.org/api/org/slf4j/helpers/NOPLogger.html can help?

tschlat: Info and Debug should also be ignore. It m

> is there a possibility to completly supress the underlying logger? because you'd have to add an immediate `return` to **every** method - also `info` and `error` etc. that seems a bit much to me....

Please see the javadoc of the class.

tschlat: > maybe this https://www.slf4j.org/api/org/slf4j/helpers/NOPLogger.html can help?

Gratefully accepted.

bbortt (approved):

christophd:
hey folks, thanks for the contribution!

I think it would also be nice to enable/disable the reporter via env var settings like HTMLReporter is doing (see [HtmlReporter](https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/org/citrusframework/report/HtmlReporter.java#L73))

Also in general for all reporters it would be good to find a way to completely remove the reporter from the list of reporters once disabled. So we do not even initialize the reporter and send log events to it once disabled. 

Reporters are initialized in [DefaultTestReporters](https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/org/citrusframework/report/DefaultTestReporters.java). So we should make this configurable and enable/disable reporters here not even initializing disabled ones.

For Spring configuration we can use
```java
@ConditionalOnProperty(
  value="citrus.default.logging.enabled", 
  havingValue = "true", 
  matchIfMissing = true)
```
on the reporter bean to not even initialize the bean based on the env var settings.

I think I will open a separate enhancement issue for this feature.

tschlat:
I implemented it like this, running into the problem, that the LogginReporter also serves as Logger for Inbound/Outbound messages.

You may want to have a look at this:
https://github.com/citrusframework/citrus/compare/main...issue/973/standard-reporter-override-TS

The point is, that the default reporters which are registered only if required, are not available as beans and are thus not autowired properly. Especially with respect to the additional interfaces, the LoggingReporter is implementing. 
I then dropped this approach.

I also already checked @ConditionalOnProperty. It would require the following dependency to be added to Citrus:

```xml
<dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-autoconfigure</artifactId>
</dependency> 
```

If we want to avoid it, I could add respective conditions and use @Conditional.

I'd be pleased to provide an MR for the conditional approach if you could shortly comment on the depencency question.

Issue #978, opened 2023-09-10
HttpMessageConverter fails on Custom HttpStatusCodes


tschlat:
**Citrus Version**
4.0 and lower

**Expected behavior**
A HttpMessage with custom status code can be converted to a citrus message.

**Actual behavior**
Conversion of Message fails because  org.citrusframework.http.message.HttpMessage.getStatusCode does not support custom codes.
It tries to resolve one of the Standard Codes from HttpStatus Enum.

**Test case sample**
The management of HttpStatus has undergone several changes in Spring, and we will provide a corresponding update in Citrus to accommodate these changes. With this, we will also provide a TestCase.

PR #981, opened 2023-09-12
fix(#978): Allow custom http status codes

Tom-29:
This pull request will fix [Issue 978](https://github.com/citrusframework/citrus/issues/978). There are a few breaking changes because `HttpMessage` uses now `HttpStatusCode` instead of the `HttpStatus` enum. I also removed the `HttpMessage.statusCode(Integer)` and the `HttpMessage.status(HttpStatus)` method because they are no longer necessary. `HttpMessage.getStatusCode()` returns now a `HttpStatusCode` instead of a `HttpStatus`.


bbortt (changes requested):
just some minor adjustments I would take to your formatting of the `if`-clauses.

bbortt (approved):

bbortt:
@Tom-29 I've squashed your commits. whenever you apply a suggestion directly in GitHub, you're creating new commits.

PR #985, opened 2023-09-15
allow WITH sql keyword for query action

### LINKED ISSUE #984: query statements may not start using the 'WITH' keyword
**Citrus Version**
3.4.0

**Expected behavior**
The statement/query, starting with the keyword 'WITH' instead of 'SELECT' will be executed.

**Actual behavior**
Exception is thrown while validating the statement, before executing.

**Test case sample**
> enter a recursive query like the following (example taken from source https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-recursive-query/)
> query is not executed because of ExecuteSqlQueryAction.validateSqlStatement

```
query(dataSource).statement("WITH RECURSIVE subordinates AS (
	SELECT
		employee_id,
		manager_id,
		full_name
	FROM
		employees
	WHERE
		employee_id = 2
	UNION
		SELECT
			e.employee_id,
			e.manager_id,
			e.full_name
		FROM
			employees e
		INNER JOIN subordinates s ON s.employee_id = e.manager_id
) SELECT
	*
FROM
	subordinates;").build();
```

Workaround:
```
testRunner.run(new ExecuteSQLQueryAction(sqlQueryActionBuilder) {
    @Override
    protected void validateSqlStatement(final String stmt) {
        if (!(stmt.toLowerCase().startsWith("select") || stmt.toLowerCase().startsWith("with"))) {
            throw new CitrusRuntimeException("Missing keyword SELECT or WITH in statement: " + stmt);
        }
    }
});
```

bbortt:
closes #984.

### DIFF
--- a/connectors/citrus-sql/src/main/java/org/citrusframework/actions/ExecuteSQLQueryAction.java
+++ b/connectors/citrus-sql/src/main/java/org/citrusframework/actions/ExecuteSQLQueryAction.java
@@ -143,14 +143,18 @@ public class ExecuteSQLQueryAction extends AbstractDatabaseConnectingTestAction 
      * @param context
      */
     protected void executeStatements(List<String> statements, List<Map<String, Object>> allResultRows, Map<String, List<String>> columnValuesMap, TestContext context) {
-        for (String stmt : statements) {
-            validateSqlStatement(stmt);
-            final String toExecute;
+        if (getJdbcTemplate() == null) {
+            throw new CitrusRuntimeException("No JdbcTemplate configured for query execution!");
+        }
 
-            if (stmt.trim().endsWith(";")) {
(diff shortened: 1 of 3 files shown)


bbortt:
@christophd I also found this while looking through the code:

https://github.com/citrusframework/citrus/blob/f21012346a1fc2c8804838085bd40a2d6ff3ec5b/connectors/citrus-sql/src/main/java/org/citrusframework/actions/ExecuteSQLQueryAction.java#L127

shall I remove it?

christophd:
> @christophd I also found this while looking through the code:
> 
> https://github.com/citrusframework/citrus/blob/f21012346a1fc2c8804838085bd40a2d6ff3ec5b/connectors/citrus-sql/src/main/java/org/citrusframework/actions/ExecuteSQLQueryAction.java#L127
> 
> shall I remove it?

yes, if not now when should we do it. please go ahead

christophd (approved):
Awesome!

Issue #993, opened 2023-09-20
Add QuarkusTest runtime support


christophd:
Quarkus is an emerging Java enterprise library that provides awesome test support with `@QuarkusTest` annotation and dev services.

Make sure that QuarkusTest and Citrus JUnit 5 extensions do work smoothly together so that Quarkus developers can make use of Citrus

bbortt:
damn. that's amazing!

PR #994, opened 2023-09-20
fix(#993): Add QuarkusTest runtime support

christophd:
- Enable Citrus support for QuarkusTest framework
- Add Citrus Quarkus test resource that takes care of Citrus resource injection and Citrus test lifecycle management as part of a Quarkus test
- Add CitrusSupport annotation to enable Citrus for a Quarkus test


christophd:
Fixes #993

PR #1001, opened 2023-09-25
#997: Replace Optional.orElse() with Optional.orElseGet()

### LINKED ISSUE #997: Lazy evaluation of Optional fallback values
Hi!

I noticed that throughout the Citrus project, there are calls to `java.util.Optional#orElse()` with arguments that could be initialized or instantiated in a lazy manner, just to name a few:
- [InputAction](https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java#L72)
- [KafkaProducer](https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaProducer.java#L75)

At those places, I think `Optional#orElseGet()` could be used instead.

I'll gladly open a PR, if you too think this could be useful.

picimako:
### Changes
- Replaced `Optional.orElse()` with `Optional.orElseGet()` where feasible, to evaluate non-static values lazily.

Fixes #997.


core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
```diff
@@ -69,7 +69,7 @@ public InputAction(Builder builder) {
         this.message = builder.message;
         this.variable = builder.variable;
         this.validAnswers = builder.validAnswers;
-        this.inputReader = Optional.ofNullable(builder.inputReader).orElse(new BufferedReader(new InputStreamReader(System.in)));
+        this.inputReader = Optional.ofNullable(builder.inputReader).orElseGet(() -> new BufferedReader(new InputStreamReader(System.in)));
```
bbortt, line 72: @christophd out of curiosity. how does citrus make sure this `inputReader` will be closed? I don't see it.

bbortt (approved):

core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
christophd: @bbortt I also do not see it being closed properly

bbortt: :octocat: @picimako would you be willing to take a look at this, too? please?

picimako: @bbortt Sure, I can look into it, and see what I can find.

bbortt: I think it's as easy as using a "try-with-resources" statement here:

```shell
diff --git a/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java b/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
index 438310067..ea2409ee0 100644
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
@@ -63,7 +63,7 @@ public class InputAction extends AbstractTestAction {
     /**
      * Default constructor.
      */
-    public InputAction(Builder builder) {
+    private InputAction(Builder builder) {
         super("input", builder);

         this.message = builder.message;
@@ -90,12 +90,9 @@ public class InputAction extends AbstractTestAction {
             display = message;
         }

-
-        try {
+        try (BufferedReader stdin = getInputReader()) {
             do {
                 logger.info(display);
-
-                BufferedReader stdin = getInputReader();
                 input = stdin.readLine();
             } while (validAnswers != null && !checkAnswer(input));
         } catch (IOException e) {
```

picimako: @bbortt Yeah, that should do it.

I also did a quick search, and I see that the same problem should be fixed in `org.citrusframework.validation.script.TemplateBasedScriptBuilder#build` too. If you have no objection, I'll open a separate PR for these two changes.

christophd: +1 on a separate PR for this. Many thanks!

bbortt (approved):

Issue #1003, opened 2023-09-26
ResourcePathTypeResolver is not threadsafe


tschlat:
**Citrus Version**
4.0

**Expected behavior**
Tests can be executed  in parallel.

**Actual behavior**
Parallel test execution fails non deterministically because auf ConcurrentModificationException.

**Stacktrace**

`
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1631)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.StreamSpliterators$WrappingSpliterator.forEachRemaining(StreamSpliterators.java:310)
	at java.base/java.util.stream.Streams$ConcatSpliterator.forEachRemaining(Streams.java:735)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at org.citrusframework.spi.ResourcePathTypeResolver.resolveAll(ResourcePathTypeResolver.java:117)
	at org.citrusframework.spi.TypeResolver.resolveAll(TypeResolver.java:75)
	at org.citrusframework.spi.TypeResolver.resolveAll(TypeResolver.java:63)
	at org.citrusframework.spi.TypeResolver.resolveAll(TypeResolver.java:52)`

tschlat:
The problem here is the new array:

`private final List<String> zipEntriesAsString = new ArrayList<>();`

Access to this array is not synchronized.

PR #1004, opened 2023-09-26
fix(#1003): Synchronize access to the zipEntries array cache

tschlat:
This is a fix for Issue 1003. Parallel test execution revealed, a synchronization issue in  ResourcePathTypeResolver.


bbortt:
@tschlat why not use `private final List<String> zipEntriesAsString = Collections.synchronizedList(new ArrayList<>());`? would also decrease the probability that this happens in the future.

tschlat:
> @tschlat why not use `private final List<String> zipEntriesAsString = Collections.synchronizedList(new ArrayList<>());`? would also reduce the probability that this happens in the future.

Fine with me.

bbortt (approved):

Issue #1014, opened 2023-10-12
Provide a way to specify a custom TestCaseRunner


tschlat:
**User story**
>As a CitrusFramework user, I want the ability to customize the TestCaseRunner to better align with my unique testing requirements.

**Additional context**
>Currently the Citrus DefaultTestCaseRunner is pretty much fixed and cannot easily be replaced, for example in the CitrusSpringExtension. I would like to be able to replace the runner to provide additional behaviour such as  retrying.

PR #1015, opened 2023-10-12
fix(#1014): Add a mechanism to provide a custom test case runner

tschlat:
Add a PR for fixing [#1014](https://github.com/citrusframework/citrus/issues/1014)

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/DefaultTestCaseRunner.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/DefaultTestCaseRunner.java
@@ -151,4 +151,18 @@ public class DefaultTestCaseRunner implements TestCaseRunner {
         this.testCase = testCase;
         this.testCase.setIncremental(true);
     }
+
+    public static class DefaultTestCaseRunnerProvider implements TestCaseRunnerProvider {
+
+        @Override
+        public TestCaseRunner createTestCaseRunner(TestContext context) {
+            return new DefaultTestCaseRunner(context);
+        }
+
+        @Override
+        public TestCaseRunner createTestCaseRunner(TestCase testCase, TestContext context) {
+            return new DefaultTestCaseRunner(testCase, context);
+        }
+
+    }
 }
--- a/core/citrus-base/src/main/java/org/citrusframework/TestCaseRunnerFactory.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/TestCaseRunnerFactory.java
@@ -1,0 +1,85 @@
+package org.citrusframework;
+
+import org.citrusframework.context.TestContext;
+import org.citrusframework.exceptions.CitrusRuntimeException;
+import org.citrusframework.spi.ResourcePathTypeResolver;
+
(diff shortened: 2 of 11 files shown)


core/citrus-base/src/main/java/org/citrusframework/TestCaseRunnerFactory.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final String DEFAULT = "default";
+
+    /** The key for a custom test case runner provider */
+    private static final String CUSTOM = "custom";
+
+    /** Test runner resource lookup path */
+    private static final String RESOURCE_PATH = "META-INF/citrus/test/runner";
+
+    /** Default Citrus test runner from classpath resource properties. Non-final to support testing.*/
+    private ResourcePathTypeResolver typeResolver = new ResourcePathTypeResolver(RESOURCE_PATH);
```
bbortt, line 39: since you're using `ReflectionTestUtils` anyway, I don't think the `final` keyword would make a difference. right?

core/citrus-base/src/main/java/org/citrusframework/TestCaseRunnerFactory.java
```diff
@@ -0,0 +1,86 @@
...
+ * The specified file must define the type of {@link TestCaseRunnerProvider} responsible for
+ * delivering the custom test case runner.
+ *
+ * @author Thorsten Schlathoelter
+ * @since 4.0
+ * @see TestCaseRunnerProvider
+ */
+public class TestCaseRunnerFactory {
+
+
```
bbortt, line 28:
```suggestion
```

core/citrus-base/src/main/java/org/citrusframework/DefaultTestCaseRunner.java
```diff
@@ -151,4 +151,18 @@ public void setTestCase(TestCase testCase) {
         this.testCase = testCase;
         this.testCase.setIncremental(true);
     }
+
+    public static class DefaultTestCaseRunnerProvider implements TestCaseRunnerProvider {
+
+        @Override
+        public TestCaseRunner createTestCaseRunner(TestContext context) {
+            return new DefaultTestCaseRunner(context);
```
bbortt, line 159: I think the constructor of `DefaultTestCaseRunner` should be `private`? otherwise we still encourage users to use it directly.

bbortt (changes requested):

core/citrus-base/src/main/java/org/citrusframework/DefaultTestCaseRunner.java
tschlat: The DefaultTestCaseRunner is directly instantiated in hundrets of testcases. Therefore I have not done this.

bbortt: ok. I mean, people _can_ still use it.

bbortt (approved):

PR #1020, opened 2023-10-17
Use try-with-resource to close BufferedReaders

picimako:
As discussed on https://github.com/citrusframework/citrus/pull/1001, I changed `BufferedReader` initializations to use try-with-resource.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
@@ -91,11 +91,9 @@ public class InputAction extends AbstractTestAction {
         }
 
 
-        try {
+        try (BufferedReader stdin = getInputReader();) {
             do {
                 logger.info(display);
-
-                BufferedReader stdin = getInputReader();
                 input = stdin.readLine();
             } while (validAnswers != null && !checkAnswer(input));
         } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


core/citrus-base/src/main/java/org/citrusframework/validation/script/TemplateBasedScriptBuilder.java
```diff
@@ -64,9 +64,8 @@ public String build() {
         StringBuilder scriptBody = new StringBuilder();
         String importStmt = "import ";
 
-        try {
+        try (BufferedReader reader = new BufferedReader(new StringReader(scriptCode));) {
```
bbortt, line 67:
```suggestion
        try (BufferedReader reader = new BufferedReader(new StringReader(scriptCode))) {
```

core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
```diff
@@ -91,11 +91,9 @@ public void doExecute(TestContext context) {
         }
 
 
-        try {
+        try (BufferedReader stdin = getInputReader();) {
```
bbortt, line 94:
```suggestion
        try (BufferedReader stdin = getInputReader()) {
```

bbortt (changes requested):
Thank you! I think these semicolons are unnecessary, if I may be so picky 😉

core/citrus-base/src/main/java/org/citrusframework/validation/script/TemplateBasedScriptBuilder.java
picimako: Yeah, nice catch. I didn't mean to add them. :)
Removed them in a new commit.

core/citrus-base/src/main/java/org/citrusframework/actions/InputAction.java
picimako: Fixed this too. :)

bbortt:
thanks so much!

Issue #1008, opened 2023-10-06
Citrus not populating varargs into docker container properties correctly


blessedchitamba:
Version - 3.4.0

**Expected behaviour**: 
When using the ```DockerExecuteAction.builder``` to create docker containers as shown in the code snippet below, methods marked as ```methodName(String... properties)``` using vararg format in the source code are supposed to properly add the properties onto the docker container. The example below shows me trying to set environment variables

```
import static com.consol.citrus.docker.actions.DockerExecuteAction.Builder.docker;
.
.
.
	$(docker().client(dockerClient)
               .create(my_image_name)
               .name("my_container")
               .env("ENV_1=True", "ENV_2=false", "ENV_3=true")
              .validateCommandResult(new CommandResultCallback<CreateContainerResponse>() {
                       @Override
                       public void doWithCommandResult(CreateContainerResponse rsp, TestContext context) {
                                       System.out.println(rsp);
                      }
            }));
```

**Actual behaviour:**
The environment variables get populated as an object reference. This is because when they are populated into the command, they are not unpacked properly. The image below is the output of a ```docker inspect container```. 
![image](https://github.com/citrusframework/citrus/assets/40063222/e83e580b-624b-4179-8831-20ad38f1fa2d)

**Suspicion:**
The method call to ```getParameters()``` here returns a ```Map<String, Object>``` and putting ```String... env``` results in an Object type being put in the map. 
![image](https://github.com/citrusframework/citrus/assets/40063222/5dc99e6f-df07-4796-91fb-7b7c576672ae)

bbortt:
thanks for the detailed report! I assume that the underlying java-docker implementation changed (we're using third party there) and you would have to change the way the `String[]` environment variables get passed here: https://github.com/citrusframework/citrus/blob/82d985f4d588d35b36718062dcc5ab35bddf9ffb/connectors/citrus-docker/src/main/java/org/citrusframework/docker/command/ContainerCreate.java#L99.

blessedchitamba:
Sure I can do that. You can give me any other info I need to know regarding that process.

bbortt:
nothing more, really. just fork the project, fix the bug, submit a PR. straight forward 😉 

note that the pipeline must pass and tests ensuring the desired behavior are always welcome. you can add me as a reviewer.

thanks very much!

blessedchitamba:
[...]

PR #1021, opened 2023-10-17
ContainerCreate varargs bug fix

blessedchitamba:
Root issue: [https://github.com/citrusframework/citrus/issues/1008]


connectors/citrus-docker/src/main/java/org/citrusframework/docker/command/ContainerCreate.java
```diff
@@ -193,6 +196,7 @@ private Capability[] getCapabilities(String addDrop, TestContext context) {
         }
 
         return capAdd;
+    	//return Arrays.stream(getVarargsParameter(addDrop, context)).map(p -> (Capability) p).collect(Collectors.toList());
```
bbortt, line 199: please remove this commented out code.

```suggestion
```

connectors/citrus-docker/src/test/java/org/citrusframework/docker/actions/DockerExecuteActionTest.java
```diff
@@ -66,10 +70,13 @@
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.verify;
+import org.mockito.ArgumentCaptor;
```
bbortt, line 79: standard imports should go before static imports, please.

connectors/citrus-docker/src/main/java/org/citrusframework/docker/command/ContainerCreate.java
```diff
@@ -89,15 +92,15 @@ public void execute(DockerClient dockerClient, TestContext context) {
         }
 
         if (hasParameter("cmd")) {
-            if (getParameters().get("cmd") instanceof Capability[]) {
+            if (getParameters().get("cmd") instanceof String[]) {
```
bbortt, line 95: I think all those if-conditions should look something like this (with an `OR`-clause):

```suggestion
            if (getParameters().get("cmd") instanceof Capability[]
                || getParameters().get("cmd") instanceof String[]) {
```

if we just remove `Capability`, as far as I see, we lose functionality. because it is part of the underlying dockerjava api: `import com.github.dockerjava.api.model.Capability;`.

bbortt (changes requested):
thanks for the contribution! I do especially like the tests 😉  however, I think we're missing functionality if we remove the `Capability`'s. please take a look at my comments.

connectors/citrus-docker/src/main/java/org/citrusframework/docker/command/ContainerCreate.java
blessedchitamba: Oh yeah I wasnt sure about those 'instanceof Capability[]' checks; they appeared to be typos for me 😅. But let me put them back

bbortt: typos don't compile 😉

blessedchitamba: PR updated

bbortt (changes requested):
cool, that looks good to me now! but I think you've messed something up while branching. would you mind rebasing to the `main`-branch and squashing your changes into one `fix: ...` commit?

> Conflicting files:
**connectors/citrus-docker/src/main/java/org/citrusframework/docker/command/ContainerCreate.java**

blessedchitamba:
Oh snap. I see..okay let me do so

blessedchitamba:
I have done the rebasing to solve conflicts. However as for the squashing, isn't that something I would do when merging? (I am mostly familiar with Bitbucket)

bbortt:
> However as for the squashing, isn't that something I would do when merging?

not necessarily, but you can of course. however now with 33 commits I think there has something gone wrong during the rebase 🐃 would you mind taking a look at it again?

blessedchitamba:
Not sure if this would be fine with you but I just did a reset hard to main, pulled latest changes, and created another new branch to which I then cherry picked my changes, then created a new PR here https://github.com/citrusframework/citrus/pull/1035

bbortt:
superseded by https://github.com/citrusframework/citrus/pull/1035.

PR #1022, opened 2023-10-17
Upgrade and Pin Netty 4.1.100.Final

apupier:
- it is a dependency of Zookeeper
- Netty 4.1.100.Final contains a fix for important CVE
- Citrus is then affected but given the context of usage of Citrus with good chance of not being vulnerable
- Zookeeper has not integrated the fixed version yet
- Pinning here will avoid having Citrus to be flagged and ease integration for consumers of Citrus


christophd (commented):
Many thanks! I think it would be good to add this into the dependencyManagement section in the root pom.xml and to introduce a new property `${netty.version}`

apupier:
> Many thanks! I think it would be good to add this into the dependencyManagement section in the root pom.xml

I tried to put it in parent pom in dependency management with an import bom and it wasn't working.
But I can try with the dependencies directly

christophd (approved):
LGTM now, thank you

apupier:
> what do you mean by the following?
> 
> > Pinning here will avoid having Citrus to be flagged and ease integration for consumers of Citrus

A new release of netty has been provided mentioning potential security issue https://netty.io/news/2023/10/10/4-1-100-Final.html . The previous versions will surely soon be flagged as affected/vulnerable which will lead Citrus to be flagged too.
Consumers of Citrus will have the same kind of problem.
For now Zookeeper has not integrated a newer version of Netty: https://github.com/apache/zookeeper/pull/2082

apupier:
argh wait, zookeeper is used in citrus-kafka too

christophd:
I see netty also in vert.x, docker-java and selenium 😄

apupier:
> vert.x, docker-java and selenium

* vert-x already rpovided a new release with upgraded netty: https://github.com/eclipse-vertx/vert.x/commits/4.4.6 so i think we need to upgrade it in another PR
* docker-java is still on an old version https://github.com/docker-java/docker-java/blob/12ecc2bcbf0b6fbfe64cc658101ebccd6a1d06d6/pom.xml#L75 (but a lot older, I hope we cna apply the sam ekkind of trick)
* Selenium has provided a new release with the fix https://github.com/SeleniumHQ/selenium/commit/8f737227121762ff3bf8748ea1bc7b9f943bd5bd so I think we need to upgrade it in another PR

apupier:
> * docker-java is still on an old version https://github.com/docker-java/docker-java/blob/12ecc2bcbf0b6fbfe64cc658101ebccd6a1d06d6/pom.xml#L75 (but a lot older, I hope we cna apply the sam ekkind of trick)

not confident for docker-java, the PR to upgrade in docker-java is failing: https://github.com/docker-java/docker-java/actions/runs/6534289853/job/17741156863?pr=2222

christophd:
Your update PRs for vert.x and selenium got merged already, thanks. How about pinning the netty version for those, too. It will avoid diverging netty versions and different versions being mixed by different `citrus-*` modules. WDYT?

Also have you tried to pin the netty version for citrus-docker module already?

apupier:
> How about pinning the netty version for those, too. It will avoid diverging netty versions and different versions being mixed by different `citrus-*` modules. WDYT?

Pinning for all will add quite a bunch of maintenance. i'm nt sure that it is worthy (but I do no think that I will be the one to do most of the maintenance for this project)

> Also have you tried to pin the netty version for citrus-docker module already?

Not yet.
Given that it means to list a longer list of netty dependency, I wanted to investigate a bit more. it seems that we could slim down from docker-java to docker-java-core and get rid of Netty dependency. At least, it compiles. I created this [PR](https://github.com/citrusframework/citrus/pull/1025) to have tests running. if it is not working, I will try to pin the version for `citrus-kubernetes` and `citrus-docker` (which are both using `docker-java`)

PR #1032, opened 2023-10-20
fix: Deactivate standard reporters

tschlat:
As it turns out, overriding Citrus reporters with this simple approach is not entirely feasible. Consequently, I have removed the respective part, and now exclusively focus on deactivating the default reporters.


core/citrus-spring/src/test/java/org/citrusframework/reporter/ReporterConfigDisabledTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.citrusframework.UnitTestSupport;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.test.context.TestPropertySource;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+
+@TestPropertySource(value = "classpath:org/citrusframework/reporter/default-logger-disabled.properties")
+public class ReporterConfigDisabledTest extends UnitTestSupport {
```
bbortt, line 15: just a side-node: that is an `*IT` imo, but I don't know if the core is already properly cleaned up. I've done it for the simulator, tho. but no objection at this point.

src/manual/reporting.adoc
```diff
@@ -27,6 +27,8 @@ public static class OverrideReporterConfiguration {
 }
 ----
 
+Note that you need to allow spring bean overriding to load this configuration as such. It's worth noting that bean overriding is not recommended by spring. Thus the following properties are provided to deactivate standard reporters as needed.
+
 In Spring, it is also possible to disable these standard reporters using the following properties:
```
bbortt, line 32: that sentence is kinda obsolete, no?

bbortt (changes requested):
thanks @tschlat. as we've already discussed, this change is ok from my POV. just one more question regarding the docs.

PR #1036, opened 2023-10-24
fix: Do not resolveAllFromJar on empty path

tschlat:
Citrus does not "mvnw verify" on windows any more.
Looking into  the following commit, one can see, that the code was not correctly adjusted for this change. 
[See 6b163eb](https://github.com/citrusframework/citrus/commit/6b163eb115814935a87cb81f8e0da5e91fb1fe6a#r130779285)

The problem occurs, when passing an empty string into resolveAllFromJar. Passing the fullPath variable solves the issue because   getFullResourcePath handles empty paths.


christophd:
Many thanks!

PR #1041, opened 2023-10-25
fix: more issues on resolving resources on Windows OS

tschlat:


core/citrus-api/src/main/java/org/citrusframework/spi/Resources.java
```diff
@@ -141,7 +141,7 @@ public boolean exists() {
 
         @Override
         public InputStream getInputStream() {
-            return ReflectionHelper.class.getClassLoader().getResourceAsStream(location);
+            return ReflectionHelper.class.getClassLoader().getResourceAsStream(location.replace("\\","/"));
```
tschlat, line 144: This is an issue, when loading the resource from a jar file.

christophd:
@tschlat thanks! looks good. is this solving all tests on Windows now?

tschlat:
I would say so. I needed to Ignore: WaitForTests and some RMI Tests. The issue with WaitForTests is probably related to our internal proxy. I had that before. 

RMI I was not sure. These tests wait for the RMI Server at the beginning and that doesn't seem to get up. I did not look into this. If you think it could have to do with this issue I can dig into it. Need to find out where and how this RMI Server is actually startet though,

christophd:
@tschlat many thanks! WaitFor and RMI flaky tests are known and should not be related to this particular run on Windows OS

PR #1044, opened 2023-10-30
fix: support loading data dictionary from xml

bbortt:
a fix for the simulator release, see [the discussion on the simulator PR](https://github.com/citrusframework/citrus-simulator/pull/188#issuecomment-1784761576).


bbortt:
I've added a test...

christophd (commented):
Looks good, thank you! 

A nitpick I have is that you are mixing TestNG unit tests with JUnit Jupiter and JUnit4 assertions. I wonder why JUnit Jupiter is even on the classpath for this module. We should not mix TestNG and JUnit Jupiter at the same time.

I would like to just use TestNG assertions as long as the whole test suite uses TestNG

bbortt:
oh, the `import static org.junit.jupiter.api.Assertions.assertThrows;`.. damn, that's a nice catch, actually. I didn't even notice! sure, I'll remove it. I know that testng does provide an almost equal one.

Issue #1046, opened 2023-11-01
Excessive classpath scanning slows down test execution


tschlat:
**Citrus Version**
4.0

**Expected behavior**
Test execution time should be comparable to previous versions

**Actual behavior**
Test execution, especially in parallel, is significant slower. The reason for this is excessive classpath scanning during type resolving via 

```java
org.citrusframework.spi.ResourcePathTypeResolver.resolveAll();
```

Especially this class is problematic:
```java
org.citrusframework.variable.SegmentVariableExtractorRegistry
```

as it is instantiated for each TestContext. So for each TestContext instantiation a classpath scan (including jars) is performed. 

The issue has been introduced with:
org.citrusframework.spi.ClasspathResourceResolver

PR #1048, opened 2023-11-02
fix(#1046): Resolve excessive class pass scanning by caching scan res…

tschlat:
Classpath scanning massively kicked in on our simulator, when a TestContext was created. Mainly due to the fact, that the org.citrusframework.variable.SegmentVariableExtractorRegistry uses the type resolver for each instance, although it could probably be a singelton.

Since the scan does a lot of file io, multiple, processes significantly slow down scenario execution time. This hold also true for parallel execution of tests, as they all try to access the limited file io resource.

I ended up implementing a caching of scanned resources in the ResourcePathTypeResolver. Although This patch solved our issues, please review carefully.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/spi/ResourcePathTypeResolver.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/spi/ResourcePathTypeResolver.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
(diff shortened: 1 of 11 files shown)


christophd (approved):
Thank you very much! Huge improvement!

core/citrus-api/src/main/java/org/citrusframework/spi/ResourcePathTypeResolver.java
```diff
@@ -232,25 +267,32 @@ private Constructor<?> getConstructor(Class<?> type, Object[] initargs) {
...
-        String path = getFullResourcePath(resourcePath);
-
-        InputStream in = ResourcePathTypeResolver.class.getClassLoader().getResourceAsStream(path);
-        if (in == null) {
-            throw new CitrusRuntimeException(String.format("Failed to locate resource path '%s'", path));
-        }
+        return resourceProperties.computeIfAbsent(resourcePath, k -> {
+            String path = getFullResourcePath(resourcePath);
+
+            InputStream in = ResourcePathTypeResolver.class.getClassLoader()
```
bbortt, line 279: I find it very hard to read, to be honest. its' probably IntelliJ auto-formatting and I know we have no `prettier` or any other rules in this project.

but, while looking through it I did at least find this line: this results in open handles.

I will follow up with another PR, reformatting this (and any other issues I might find).

bbortt: see 3443c8bb94f9c181d0041f8afaed43c73aec2697 in https://github.com/citrusframework/citrus/pull/1044.

PR #1053, opened 2023-11-09
fix(#1052): Provide ConversionService configuration as extra Configuration class

### LINKED ISSUE #1052: Spring custom converters are not automatically registered
**Citrus Version**
4.0

**Expected behavior**
Custom `org.springframework.core.convert.converter.Converter´ should automatically be registered with the spring conversion service, as this was done in Citrus 3.0.

**Actual behavior**
Custom converters are not registered, due to the fact, that CirtusSpringConfiguration registers a `DefaultConversionService` which is as such not capable to register custom converters. This registration prevents spring from creating it's own extendable conversion service.

**Additional context**
@SpringBootTests set up a context with a proper extendable conversion service. It seems like Spring only applications need to register an own ConversionService. Therefore in Citrus, ConversionService configuration should be separated into an own Configration class, which can be added as needed.

tschlat:
Fixes #1052

### DIFF
--- a/core/citrus-spring/src/main/java/org/citrusframework/config/CitrusConversionServiceConfiguration.java
+++ b/core/citrus-spring/src/main/java/org/citrusframework/config/CitrusConversionServiceConfiguration.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.citrusframework.config;
+
+import java.util.Set;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.support.ConversionServiceFactoryBean;
+import org.springframework.core.convert.converter.Converter;
+
+/**
+ * Optional configuration providing a {@link org.springframework.core.convert.ConversionService} with Citrus-related converters.
+ * <p>
+ * This configuration is necessary only when using Spring without Spring Boot. Spring Boot includes a standard conversion service
+ * that automatically detects and uses relevant converters by default.
+ *
+ * @author Thorsten Schlathoelter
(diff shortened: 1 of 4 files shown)


christophd:
Hmm, I do not think that this is limited to TestNG. We need to then find a proper fix for adding the Citrus custom converters and allow users to add their custom converters at the same time

bbortt (approved):

bbortt:
@tschlat careful when going further with this. I've adjusted the copyright settings. you have to reset your local branch to the remote first.

christophd (changes requested):
I think only setting this for TestNG does not solve it for other runtimes such as JUnit Jupiter.

What is the recommended way of adding custom converters to a Spring application context other than setting the conversion service?

tschlat:
I delved deeper into this matter, and from my understanding, Spring Boot automatically registers its own ConversionService if no other ConversionService is registered by the application. This service is instantiated in org.springframework.boot.context.properties.ConversionServiceDeducer. The default service is designed to discover other Converter beans, which can be either registered as beans within Configuration classes or annotated with @Component. However, if the application registers its own ConversionService, it becomes responsible for handling the inclusion of other converters.

It appears that Spring standalone does not offer a similar built-in ConversionService by default, at least not in the configuration commonly used in Citrus TestNG tests. Consequently, I suggest segregating the Citrus ConversionService configuration into a distinct Configuration class, as implemented in this pull request. This way, users can decide whether a distinct ConversionService should be registered by Citrus or not. For Citrus TestNG Tests I have added this configuration. For SpringBoot tests the CitrusSpringConfiguration is sufficient, as it now registers the Citrus specific converters as beans. 

I have adjusted the PR a litte to clarify this.

It's worth noting that with the current Citrus implementation, registering application-specific converters is possible. This involves employing a BeanFactoryPostProcessor and manually registering the application-specific converters with the Citrus ConversionService. Therefore, this change isn't crucial but provides users with the flexibility to make such decisions.

tschlat:
Furthermore, I've introduced the CitrusResourceEditor along with a corresponding org.springframework.beans.factory.config.CustomEditorConfigurer. This alleviates the need to manually register the ConversionServiceConfiguration. The editor facilitates the conversion from a String to a Resource.

tschlat:
> I think only setting this for TestNG does not solve it for other runtimes such as JUnit Jupiter.
> 
> What is the recommended way of adding custom converters to a Spring application context other than setting the conversion service?

A conversion service is OK but one has to keep in mind, that there might be side effects with other conversion services and converters. A more robust approach to offer String to Citrus Resource conversion is through a PropertyEditor, seamlessly integrating with Spring. I have added one in this PR.

christophd:
Ah, I was missing the PropertyEditor. Now looks good to me, thank you for solving this issue!

bbortt:
damn, nice deep dive :) well done.

PR #1065, opened 2023-11-19
fix: cookie on http response with status code

bbortt:
this has been around a while. I wonder why this has popped up now @tschlat. but it indeed seems to be an issue, report in https://github.com/citrusframework/citrus-simulator/issues/218.


christophd (approved):
I have no idea why cookie handling could have been added into that else case. 🤷 

Looks good as it is right now

PR #1067, opened 2023-11-20
fix: data dictionary resolve for classpath resources

bbortt:
why did nobody tell me that `ClasspathResource`'s don't have a `File`? 😉 

fixes https://github.com/citrusframework/citrus-simulator/pull/219.


bbortt:
wsdl-sample still fails in citrusframework/citrus-simulator. will have to check if this is related.

Issue #1066, opened 2023-11-20
Improvements for `JmsEndpointConfigParser` and `JmsSyncEndpointConfigParser`


picimako:
Hi!

I'm looking into how `JmsEndpointConfigParser` and `JmsSyncEndpointConfigParser` works, so that I could implement some IntelliJ inspections for their respective config annotations, and I noticed a few things that could be improved in and regarding them:

- I cannot see a way in which the exception *"Required connection-factory is missing for jms configuration"* would be thrown. Since `connectionFactory` has a default value, `if (!StringUtils.hasText(connectionFactory))` will never be true.
```java
//connectionFactory
String connectionFactory = "connectionFactory"; //default value

if (StringUtils.hasText(annotation.connectionFactory())) {
    connectionFactory = annotation.connectionFactory();
}

if (!StringUtils.hasText(connectionFactory)) {
    throw new CitrusRuntimeException("Required connection-factory is missing for jms configuration");
}
```

- Regarding the exception *""When providing a jms-template, none of connection-factory, destination, or destination-name should be provided"*:
  - This is thrown in the else branch of `StringUtils.hasText(destination) || StringUtils.hasText(destinationName)`, so in that else branch both `destination` and `destinationName` are empty, thus checking for `StringUtils.hasText(destination) || StringUtils.hasText(destinationName)` there seems unnecessary.
  - I think the destination and destination name checks should be removed from the else branch, and the `jmsTemplate` text check/branch should be copied to the `if (StringUtils.hasText(destination) || StringUtils.hasText(destinationName))` block of the condition, **something** like this:

```java
if (StringUtils.hasText(destination) || StringUtils.hasText(destinationName)) {
    //...

    if (StringUtils.hasText(jmsTemplate)) {
        //...
    }
} else if (StringUtils.hasText(jmsTemplate)) {
    if (StringUtils.hasText(annotation.connectionFactory())) {
        throw new CitrusRuntimeException("When providing a jms-template, none of " +
                "connection-factory, destination, or destination-name should be provided");
    }

    builder.jmsTemplate(referenceResolver.resolve(jmsTemplate, JmsTemplate.class));
}
```

- The exception branches of the conditions are not covered by unit tests, so those could be added.

Let me know what you think.

bbortt:
> I cannot see a way in which the exception "Required connection-factory is missing for jms configuration" would be thrown. Since connectionFactory has a default value, if (!StringUtils.hasText(connectionFactory)) will never be true.

[...]

PR #1068, opened 2023-11-21
#1066: Improve validation and test coverage for JmsEndpointConfigParser and JmsSyncEndpointConfigParser

picimako:
### Changes
- Removed the `connectionFactory` related exception throwing.
- Move exception throwing in case of jms template and destination/destination name combination to the proper `if` block.
- Added unit tests for the exception flows of both parser classes.


christophd (approved):
Good work! Many thanks!

PR #1069, opened 2023-11-21
chore: citrus resource wrapper for classpath resources

bbortt:
after https://github.com/citrusframework/citrus/pull/1067 I figured that it'd be best to centralize property loading, so errors can be minimalized in the future.
additionally, classpath resources seem to require special treatment in the citrus/spring resource wrapper. that has been done poorly so far.

with this the [citrusframework/citrus-simulator builds on `4.1.0-SNAPSHOT`](https://github.com/citrusframework/citrus-simulator/pull/219). I think it would be fair to release `4.0.2` as soon as possible @christophd.


bbortt:
I've added a test as well.

christophd (approved):
Awesome! Many thanks

Issue #1071, opened 2023-11-22
ClasspathResourceResolver does not support loading resources from fat jars


tschlat:
**Citrus Version**
4.0

**Expected behavior**
Citrus should be able to run from a Spring Boot application.

**Actual behavior**
Citrus does not resolve all required components like validators etc. These are loaded by classpath scanning and this fails on jars, that are nested in a fat jar, like in Spring Boot application.

**Test case sample**
We run a Citrus Simulator in a K8S environment and we have plenty of issues, because components are not properly resolved. Remote debugging provided insight into this issue.

bbortt:
@tschlat I think that issue has been solved. report back if that is not the case.

Issue #876, opened 2022-07-27
migrate tests to junit jupiter


bbortt:
as the title states.. migrate the internal tests from testng to junit jupiter.

christophd:
Also refactor test classes to not expose as public classes as mentioned in https://github.com/citrusframework/citrus/pull/954#discussion_r1257527417

PR #1072, opened 2023-11-22
fix(#1071): resolve resources in fat jars

tschlat:
I am pretty sure this works, as i tested it locally by remote debugging a spring boot application. So feel free to review. 
However, I have a huge amount of tests waiting to be executed against a real world simulator. Before I finish this I would like to wait for these results, to make sure I have not left out something.


christophd (approved):
This looks good. Thanks a lot

core/citrus-api/src/main/java/org/citrusframework/spi/ResourcePathTypeResolver.java
```diff
@@ -191,8 +189,11 @@ private String toCacheKey(String path, String property, String keyProperty) {
 
     private Stream<Path> resolveAllFromJar(String path) {
         String rootAsString = ResourcePathTypeResolver.class.getProtectionDomain().getCodeSource().getLocation().toString();
+
         ClassLoader classLoader = ObjectHelper.assertNotNull(ResourcePathTypeResolver.class.getClassLoader());
-        if (rootAsString.endsWith(".jar") && !rootAsString.matches(".*" + File.separator + "citrus-api-\\d+\\.\\d+\\.\\d+(-.*)?\\.jar")) {
+        if (rootAsString.matches(".*jar(!/)?") &&
+            !rootAsString.replace("\\", "/")
+                .matches(".*/citrus-api-\\d+\\.\\d+\\.\\d+(-.*)?\\.jar")) {
```
bbortt, line 196: hm. I remember @Tom-29 working on this section too. has this resolved his issue as well?

core/citrus-api/src/test/java/org/citrusframework/spi/ClassPathResourceResolverTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Set;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
```
bbortt, line 13: you're mixing JUnit with TestNG. I think this test won't run with Maven, because it almost for sure has no JUnit runner. please double-check this.

bbortt (commented):
well, PR has already been merged. but I am pretty sure that this test will never run! please check this and follow up with another PR.

also, please don't do this while being ill! get well soon @tschlat 😉

tschlat:
Good catches. 
https://github.com/citrusframework/citrus/pull/1073

core/citrus-api/src/test/java/org/citrusframework/spi/ClassPathResourceResolverTest.java
bbortt: it is planned to migrate: https://github.com/citrusframework/citrus/issues/876. we've postponed it from 3.0 however, because it is a huge task!

PR #1073, opened 2023-11-22
Draft: Cleanup: Cclasspath resource resolving in fat jars

tschlat:
A little cleanup.


bbortt:
@tschlat is this still an issue?

PR #1075, opened 2023-11-26
Feature: Sharded Test Execution

bbortt:
I'd like to introduce a sharded test loader. It allows splitting test executions upon multiple runners, for parallel execution.

Use case: We have a huge testing-base in one project, with around 2000 citrus tests. Whenever we execute this, it takes up to 20 minutes. With this feature, we could run these tests on multiple workers, in parallel. Expected gain is roughly `[TIME] * 1 / 2` per runner.

Configuration is possible from environment variables, properties or Java DSL. See the updated documentation for more information.

I think it's a good feature that once more supports the state-of-the-art look of citrus.


christophd:
I wonder if someone could achieve the same with just using the unit test framework capabilities such as `@TestFactory` in JUnit Jupiter or `@DataProvider` in TestNG. And use the Maven surefire/failsafe parallel testing capabilities

bbortt:
> And use the Maven surefire/failsafe parallel testing capabilities

that is not what this is for.. this is for sharded **runners** (entirely encoppled physical or virtual machines).

> I wonder if someone could achieve the same with just using the unit test framework capabilities such as @TestFactory in JUnit Jupiter or @DataProvider in TestNG.

I wondered the same, but decided to integrate it into the JUnit (and TestNG coming soon) support of citrus.. that way it should be possible. it's already doing the same, but without the sharded part.

runtime/citrus-xml/src/test/java/org/citrusframework/integration/ShardedXmlTestLoaderIT.java
```diff
@@ -0,0 +1,41 @@
...
+import org.testng.annotations.Test;
+
+import static org.citrusframework.common.ShardedTestLoaderAndExecutor.XML_SHARDED;
+
+@Test
+public class ShardedXmlTestLoaderIT extends TestNGCitrusSupport {
+
+    @CitrusTestSource(type = XML_SHARDED, sources = {"classpath:org/citrusframework/integration"})
+    public void loadFromSources() {
+        // TODO: Check if test has been executed
```
bbortt, line 14: question @christophd, how would you do this? how can I check how many tests have actually been executed? e.g. here it would be one, in the recursive tests it would be three?

christophd: maybe add a custom test reporter (or test listener) that counts the number of executed tests. and after the test run verify the tests that reported success/failure in the custom reporter.

bbortt:
> I wonder if someone could achieve the same with just using the unit test framework capabilities such as @TestFactory in JUnit Jupiter or @DataProvider in TestNG. And use the Maven surefire/failsafe parallel testing capabilities

@christophd that comment of you made me think a lot. I actually wanted to come back to this much earlier, but didn't find time. anyway, here it is. I've removed most of the previous changes and merged everything into kind of a "utility class" that takes care of sharding a given stream. this way the sharding is test framework independent. would be great if you'd review this PR once more.

christophd (approved):
Looking good, thank you! I like really the fact that this is pluggable as a wrapper to the JUnit Jupiter dynamic test factory. Really nice!

Just a nitpicking about naming of system properties and env vars

core/citrus-base/src/main/java/org/citrusframework/sharding/ShardingConfiguration.java
```diff
@@ -0,0 +1,194 @@
...
+ *   </li>
+ *   <li>
+ *       Optionally, set a seed for shuffling test cases using <code>ORG_CITRUSFRAMEWORK_SHARDING_SEED</code>
+ *       or <code>org.citrusframework.sharding.seed</code>. The total number of shards will be used as see by default.
+ *   </li>
+ * </ul>
+ *
+ * @see Shard
+ */
+public final class ShardingConfiguration {
```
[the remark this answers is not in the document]
bbortt: true, I had this in mind but forgot about those after copying it. thanks for the remark!

bbortt:
> I like really the fact that this is pluggable as a wrapper to the JUnit Jupiter dynamic test factory.

thanks! I did a quick search through the internet and saw that TestNG too has an `@Factory` annotation, but we haven't made use of it yet.

I would love to follow up with an implementation similar to `org.citrusframework.junit.jupiter.CitrusTestFactorySupport`, but for TestNG in another PR. and then integrate this one with the sharded wrapper too. that makes the package complete, for both runtimes.

Issue #1076, opened 2023-11-27
Documentation example not correct


hmmlopez:
**Citrus Version**
4.0.2

**Expected behavior**
In the [cucumber](https://citrusframework.org/citrus/reference/4.0.2/html/index.html#runtimes-cucumber) documentation there is an error in the example of the EchoSteps.java, in the last part (assertion) the verify_return method should be:
```
    @Then("^the service should return: \"([^\"]*)\"$")
    public void verify_return(final String body) {
        runner.then(
                receive("echoEndpoint")
                    .message()
                    .type(MessageType.PLAINTEXT)
                    .body(body));
    }
```
instead of
```
    @Then("^the service should return: \"([^\"]*)\"$")
    public void verify_return(final String body) {
        runner.then(
                receive("echoEndpoint")
                    .message()
                    .type(MessageType.PLAINTEXT)
                    .body("You just said: " + body));
    }
```

It would be also nice to explain how to create an Instance of that **echoEndpoint**. 
When using spring I had to create a bean of:
```
    @Bean
    fun echoEndpoint() : Endpoint {
        return DirectEndpoint().apply {
            endpointConfiguration.queue = DefaultMessageQueue("echoEndpoint")
        }
    }
```
But it took me a long time to figure it out.

bbortt:
would you be willing to submit these valuable improvements in a PR? 🚀

christophd:
sure, this is the place to look for the docs source files: https://github.com/citrusframework/citrus/tree/main/src/manual

hmmlopez:
I've created [this](https://github.com/citrusframework/citrus/pull/1080) pull request.

PR #1080, opened 2023-12-01
Removed extra text in the verification message.

hmmlopez:
The verification step should contain the expected message and not add extra text.


bbortt:
closes https://github.com/citrusframework/citrus/issues/1076.

hmmlopez:
Sure, I could but this is my first PR and wanted to see how it gets through first.

Issue #1089, opened 2024-01-08
kubernetes client: allow config with authentication token


valisol:
**User story**
As a tester, I want to create a xml test using the kubernetes client with an authentication token so that i can use the kubernetes remote API.

**Additional context**
The Kubernetes client has already been implemented, currently facilitating connections solely through a username and password.

PR #1091, opened 2024-01-09
feat(#1089): kubernetes client config with auth token

valisol:
- Allow configuration with new oauthToken
- Update Tests
- Update Documentation

### DIFF
--- a/connectors/citrus-kubernetes/src/main/java/org/citrusframework/kubernetes/config/annotation/KubernetesClientConfigParser.java
+++ b/connectors/citrus-kubernetes/src/main/java/org/citrusframework/kubernetes/config/annotation/KubernetesClientConfigParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2016 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,7 +22,9 @@ import org.citrusframework.kubernetes.client.KubernetesClient;
 import org.citrusframework.kubernetes.client.KubernetesClientBuilder;
 import org.citrusframework.kubernetes.message.KubernetesMessageConverter;
 import org.citrusframework.spi.ReferenceResolver;
-import org.citrusframework.util.StringUtils;
+
+import static org.citrusframework.kubernetes.config.CredentialValidator.isValid;
+import static org.citrusframework.util.StringUtils.hasText;
 
 /**
  * @author Christoph Deppisch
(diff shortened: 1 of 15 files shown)


src/manual/endpoint-kubernetes.adoc
```diff
@@ -69,6 +69,8 @@ In case you need to set the client configuration explicitly on your environment
               object-mapper="objectMapper"/>
 ----
 
+*Note*: You have the option to set an `oauthToken` for the client instead of using `username` and `password`.
```
bbortt, line 72: I wonder if we could check if it is **either** the token **or** username and password.. a check would probably make sense; [here](https://github.com/citrusframework/citrus/blob/deac0ef52e2735ac46ce6dcc9753f34d8466a77a/connectors/citrus-kubernetes/src/main/java/org/citrusframework/kubernetes/client/KubernetesClientBuilder.java#L41).

bbortt:
the dependency version we're using is quite old..

```xml
<k8s.client.version>1.4.34</k8s.client.version>
<k8s.model.version>1.0.65</k8s.model.version>
```

version 6 of the fabric8 client is out by now, I think. would you mind checking if the upgrade would cause big troubles? maybe we can just update the numbers without any big migrations.

I am not sure if a this-old-client is even useful with "modern" kubernetes platforms, that's why I am asking.

[link to `pom.xml`](https://github.com/citrusframework/citrus/blob/main/pom.xml#L971).

christophd:
yes we need to update to new fabric8 client version, but unfortunately this is not an easy task as it lots of breaking changes in the client API.

src/manual/endpoint-kubernetes.adoc
valisol: check was added to KubernetesClientConfigParser

connectors/citrus-kubernetes/src/main/resources/org/citrusframework/schema/citrus-kubernetes-config.xsd
```diff
@@ -30,6 +30,7 @@
       <xs:attribute name="version" type="xs:string"/>
       <xs:attribute name="username" type="xs:string"/>
       <xs:attribute name="password" type="xs:string"/>
+      <xs:attribute name="oauthToken" type="xs:string"/>
```
bbortt, line 33: the `KubernetesClientParser` misses this property for copying.

connectors/citrus-kubernetes/src/main/java/org/citrusframework/kubernetes/config/annotation/KubernetesClientConfigParser.java
```diff
@@ -34,38 +34,58 @@ public class KubernetesClientConfigParser implements AnnotationConfigParser<Kube
     public KubernetesClient parse(KubernetesClientConfig annotation, ReferenceResolver referenceResolver) {
         KubernetesClientBuilder builder = new KubernetesClientBuilder();
 
-        if (StringUtils.hasText(annotation.url())) {
-            builder.url(annotation.url());
-        }
+        if (isClientWithAuthToken(annotation) || isClientWithUsernameAndPassword(annotation) || isClientDefaultConfigured(annotation)) {
```
bbortt, line 37: to keep the change smaller and the code more readable, I suggest 
```suggestion
if (!isClientWithAuthToken(annotation) && !isClientWithUsernameAndPassword(annotation) && !isClientDefaultConfigured(annotation)) {
            throw new IllegalArgumentException("Parameters not set correctly - check if either an oauthToke or password and username is set");
}
```

the, continue with the old code.

also, this sanity check does not take invocation from XML into account, as far as I see. it does only handle the annotatoin-case, that's why the parameter is the annotation itself: `KubernetesClientConfig `.

connectors/citrus-kubernetes/src/main/resources/org/citrusframework/schema/citrus-kubernetes-config.xsd
valisol: The parser contains already the property

bbortt:
awesome job @valisol! I've allowed myself to do some "contributor stuff" (things I wouldn't expect a first-time contributor to do), like updating all copyrights and most noteable; centralizing the rules in a `CredentialValidator`. I've then unit-tested the validation according to the following rules (see javadoc):

Validates the given credentials based on the following rules. Returns {@code true} if and only if none of the following conditions are met:
<ul>
  <li>username is set AND oauthToken is set</li>
  <li>password is set AND oauthToken is set</li>
  <li>username is <b>not</b> set AND password is set</li>
</ul>

All other combinations are valid. This makes it also compatible with the existing API, removing the need to specify an empty password.

bbortt (approved):

PR #1093, opened 2024-01-15
feat: xml validation add fail if no schema found attribut

Tom-29:
This PR will add a new feature to the xml validation, which allows you to set a new property called `org.citrusframework.xml.fail.if.no.schema.found`. This is useful if you have a xmlns in the xml file which isn't in the XsdSchmaRepository. If the new property is set to false it won't fail there (`schemaRepository.canValidate(doc)`). The default is true and therefore the same behavior as before.


validation/citrus-validation-xml/src/main/java/org/citrusframework/validation/xml/schema/XmlSchemaValidation.java
```diff
@@ -33,14 +24,45 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static java.lang.Boolean.parseBoolean;
+import static java.util.Optional.ofNullable;
+
 public class XmlSchemaValidation implements SchemaValidator<XmlMessageValidationContext> {
 
+    public static final String FAIL_IF_NO_SCHEMA_FOUND_PROPERTY_NAME = "org.citrusframework.xml.fail.if.no.schema.found";
```
bbortt, line 42: I think `citrus.xml` is enough, because we've other properties that start this way. e.g. `citrus.yaml` or `citrus.json` etc.

how about `citrus.xml.no.schema.found.strategy`? that could be an `enum` with the values `IGNORE` and `FAIL`. it leaves the option to add other behaviors in the future, without breaking much.

validation/citrus-validation-xml/src/main/java/org/citrusframework/validation/xml/schema/XmlSchemaValidation.java
```diff
@@ -146,4 +176,38 @@ private void validateSchema(Message message, TestContext context, XmlMessageVali
...
+    private static boolean extractEnvOrProperty(SystemProvider systemProvider, String envVarName, String fallbackPropertyName, boolean defaultValue) {
+        return parseBoolean(extractEnvOrProperty(systemProvider, envVarName, fallbackPropertyName).orElse(String.valueOf(defaultValue)));
+    }
+
+    private static Optional<String> extractEnvOrProperty(SystemProvider systemProvider, String envVarName, String fallbackPropertyName) {
+        return systemProvider.getEnv(envVarName)
+                .or(() -> systemProvider.getProperty(fallbackPropertyName));
+    }
+
+    public static final class SystemProvider {
```
bbortt, line 202:
```suggestion
    static final class SystemProvider {
```

validation/citrus-validation-xml/src/test/java/org/citrusframework/validation/xml/schema/XmlSchemaValidationTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertThrows;
+
+public class XmlSchemaValidationTest {
+
+    @DataProvider(name = "booleans")
+    public Object[] dataProvider() {
+        return new Object[] {
+                true,
+                false
```
bbortt, line 35: if this was booleans, I would always test "other values" (e.g. exactly this value) to see what the behavior would be. but since I would change this to an `enum`, it is another story entirely.

validation/citrus-validation-xml/src/test/java/org/citrusframework/validation/xml/schema/XmlSchemaValidationTest.java
```diff
@@ -0,0 +1,110 @@
...
+        };
+    }
+
+    @Test
+    public void configurationFromEnvTakesPrecedenceOverProperties() {
+        XmlSchemaValidation.SystemProvider systemProviderMock = mock(XmlSchemaValidation.SystemProvider.class);
```
bbortt, line 41: you can still have this as an `@Mock`, when using the `MockitoAnnotations#openMocks` method.

bbortt (changes requested):

bbortt (approved):

PR #1097, opened 2024-01-17
fix: nested jar loading with relative URL

bbortt:
turns out we have jars that use relative URLs instead of absolute ones @tschlat. the `substring`-invocation then fails, respectively returns a `null`-value for `JarEntry jarEntry` which makes the subsequent call to `jarFile.getInputStream(jarEntry)` fail.


tschlat:
thx

tschlat (commented):
lgtm

Thanks for cleaning up

Issue #1101, opened 2024-01-18
JsonTextMessageValidator should ignore order of json array items


novarx:
**Citrus Version**

4.1.0

**Context**

[`JsonTextMessageValidator#validateMessage`](https://github.com/citrusframework/citrus/blob/main/validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java)

**Expected behavior**

The validation of a given json array, compared to a control json, should ignore the order of array items.
i.e. the follwing "recived-json"

```json
{ "books": ["book-c", "book-b", "book-a"] }
```
should throw no ValidationException when compared with the follwing "control-json"

```json
{ "books": ["book-a", "book-b", "book-c"] }
```

**Actual behavior**

The item ordering is always asserted, even when not in `strict` mode.

**Test case sample**

See Pull Request (#1102)

bbortt:
fixed in https://github.com/citrusframework/citrus/pull/1102.

PR #1102, opened 2024-01-18
fix(#1101): Make JsonTextMessageValidator ignore the order of JsonArrays

novarx:
See #1101 

@bbortt looking forward to your review / opinion =)


validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
```diff
@@ -0,0 +1,272 @@
...
+        var jsonAssertion = new JsonAssertion(
+                "{\"index\": {\"something\": [0]} }",
+                "{\"index\": {\"anything\": [55, 66, 77]} }",
+                Set.of("$.index['anything'][*]")
+        );
+        var validationItem = toValidationItem(jsonAssertion);
+
+        fixture = new JsonElementValidator(NOT_STRICT, context, jsonAssertion.ignoreExpressions);
+
+        // TODO that shouldn't fail, should it?
```
novarx, line 252: On second thought... 

`$.index['anything'][*]` should ignore all the children of `anything`, but the entry itself should be there nonetheless. Will update the test accordingly.

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
```diff
@@ -56,190 +46,74 @@
...
-     */
-    private List<JsonSchemaRepository> findSchemaRepositories(TestContext context) {
-        return new ArrayList<>(context.getReferenceResolver().resolveAll(JsonSchemaRepository.class).values());
-    }
-
     /**
      * Validates JSON text with comparison to expected control JSON object.
      * JSON entries can be ignored with ignore placeholder.
+     * <p>
+     * TODO is that method still needed?
```
novarx, line 96: It's nowhere used, but could still be a braking change - any thoughts?

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
```diff
@@ -56,190 +46,74 @@
...
-        }
+        var entryControl = new JsonElementValidatorItem<>(elementName, receivedJson, controlJson);
+        new JsonElementValidator(strict, context, validationContext.getIgnoreExpressions()).validate(entryControl);
     }
 
     /**
      * Checks if given element node is either on ignore list or
      * contains @ignore@ tag inside control message
+     * <p>
+     * TODO is that method still needed?
```
novarx, line 114: It's nowhere used, but could still be a braking change - any thoughts?

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
```diff
@@ -0,0 +1,135 @@
+package org.citrusframework.validation.json;
```
bbortt, line 1: please add the missing copyright for 2024. also, all touched classes should be updated to copyright years "up to 2024" (`...-2024`).

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+
+    private void validateJSONArray(JsonElementValidator validator, JsonElementValidatorItem<?> control) {
+        var arrayControl = control.ensureType(JSONArray.class);
+        if (strict) {
+            validateSameSize(control.getJsonPath(), arrayControl.expected, arrayControl.actual);
+        }
+        for (int i = 0; i < arrayControl.expected.size(); i++) {
+            if (!isAnyValidItemInActualArray(validator, arrayControl, i)) {
```
bbortt, line 97: so just to be clear; strict mode does **not** strictly validate the order of elements - right? I think that is kinda okay, since JSON in general has no such restriction. on the other hand, someone might want to test it e.g. when the return something ordered (paginated) from their APIs. we would not allow them to do so with the current implementation. any thoughts?

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
bbortt: I think just deprecate it for removal and leave an `@deprecated use .. instead` comment. this way we can remove it within one of the next few releases.

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
bbortt: I agree.

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
```diff
@@ -0,0 +1,272 @@
...
+        assertThatThrownBy(() -> fixture.validate(validationItem))
+                .isInstanceOf(ValidationException.class)
+                .hasMessage("Missing JSON entry, expected 'anything' to be in '[something]'");
+    }
+
+    private static JsonElementValidatorItem<Object> toValidationItem(JsonAssertion jsonAssertion) {
+        return JsonElementValidatorItem.parseJson(DEFAULT_PERMISSIVE_MODE, jsonAssertion.actual, jsonAssertion.expected);
+    }
+
+    public record JsonAssertion(
```
bbortt, line 262: doesn't need to be `public`, right?

bbortt (changes requested):
couldn't find much, that is a very good addition and refactoring. thank you!

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
novarx: Yeah that's true, I also thought about that...

Options are:
1. Always ignore order
2. Ignore order if `strict = false` and enforce order if `strict = true`
3. Make `JsonElementValidator` more configurable with properties 
4. Make `JsonElementValidator` more configurable by easily allowing overriding of the element-validations (in the refactoring process, I thougt of a [Chain of Responsibility](https://refactoring.guru/design-patterns/chain-of-responsibility) where each link is one element validation... but found that too sophisticated for the moment)

(I fancy Option 1 or 2)

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
novarx: Added `@deprecated use .. instead`

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidatorItem.java
```diff
@@ -0,0 +1,164 @@
...
+        this.name = name;
+        this.index = null;
+        this.actual = actual;
+        this.expected = expected;
+    }
+
+    /**
+     * @return {@link JsonElementValidatorItem#actual} as string or null
+     */
+    public String actualString() {
```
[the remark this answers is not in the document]
novarx: changed it to `actualAsStringOrNull()`

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
bbortt: so, option one is now implemented. I think that is acceptable. if someone really wants to check the order, he can easily implement a custom validator.

bbortt (approved):

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
```diff
@@ -56,225 +46,85 @@
...
+     * @deprecated use {@link JsonElementValidator#validate} instead
+     *
      * @param controlKey
      * @param controlValue
      * @param receivedJson
      * @param ignoreExpressions
      * @param readContext
      * @return
      */
+    @Deprecated(forRemoval = true)
```
christophd, line 124: If this method is not called anymore in the new validation logic we should rather remove it now. This is because somebody might have customized this method and if it is not called anymore it may be more confusing. Rather break the API and let users adjust accordingly.

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
```diff
@@ -56,225 +46,85 @@
...
+     * @param receivedJson      the received JSON text object.
+     * @param controlJson       the expected control JSON text.
      * @param validationContext the JSON message validation context.
-     * @param context the current test context.
-     * @param readContext the JSONPath read context.
+     * @param context           the current test context.
+     * @param readContext       the JSONPath read context.
      */
-    @SuppressWarnings("rawtypes")
+    @Deprecated(forRemoval = true)
```
christophd, line 105: If we deprecate we should call the method as part of the new logic. Otherwise when the method is not called I would rather remove and break the API. Then users that have added customizations to this method get noticed when updating and can adjust accordingly.

@bbortt WDYT?

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
```diff
@@ -56,225 +46,85 @@
...
-        } catch (ParseException e) {
-            throw new CitrusRuntimeException("Failed to parse JSON text", e);
+        if (!hasText(controlJsonText)) {
+            logger.debug("Skip message payload validation as no control message was defined");
+            return;
+        } else if (!hasText(receivedJsonText)) {
+            throw new ValidationException("Validation failed - expected message contents, but received empty message!");
         }
 
+        new JsonElementValidator(strict, context, validationContext.getIgnoreExpressions()).validate(
```
christophd, line 86: How about sharing the instance of JsonElementValidator in this class as a private member. Users may want to set a customized implementation of JsonElementValidator then.

We can pass the parameters to the JsonElementValidator#validate method then (e.g. context and validationContext) so JsonElementValidator is stateless

christophd (commented):
Many thanks for this contribution! I have some comments on the deprecated methods. Please have a look and let me know if you have questions.

Special thanks for refactoring the unit tests! I can not verify that all test case variants are covered as before (or even more than that) but I trust you 😃. Thanks a lot!

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
bbortt: that argumentation is valid. it support it 😉

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
novarx: Removed that method completely 👍

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonTextMessageValidator.java
novarx: I've introduced a `JsonElementValidator.Provider` to configure `JsonTextMessageValidator` with a custom element validator - hope that's fine?

novarx:
I accidentally closed the PR, didn't mean to - sorry 😅

validation/citrus-validation-json/pom.xml
```diff
@@ -72,5 +72,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.25.1</version>
```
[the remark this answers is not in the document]
novarx: Done!

bbortt (changes requested):
almost.. - sorry 😅

novarx:
Thanks for the reviews!

@christophd I've double-checked, that I only deleted testcases in `JsonTextMessageValidatorTest` that I moved to `JsonElementValidatorTest`. Further, I've added some array-ordering-tests.

bbortt:
finally.. god, that build 🤣

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
novarx: seems to be a problem - addressed in https://github.com/citrusframework/citrus/pull/1155

PR #1104, opened 2024-01-19
allow configuration of xml validator via constructor

bbortt:
in addition to https://github.com/citrusframework/citrus/pull/1093, so it's possible to override the validation style from children/callees.


christophd (approved):
LGTM

Issue #1100, opened 2024-01-18
Resources.UrlResource#exists returns false if file is empty


tschlat:
**Citrus Version**
4.x

**Expected behavior**
UrlResource#exists should return true, if the file exists but is empty.

**Actual behavior**
UrlResource#exists returns false, if the file exists but is empty.

**Test case sample**
The implementation tests the content length which is 0, if the file is empty

PR #1114, opened 2024-01-29
fix(#1100): empty file resource returns true on exist

tschlat:
Fixes issue #1100:
An UrlResource pointing to an existing but empty file must return true on exist.

Also added a couple of tests.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/spi/Resources.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/spi/Resources.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2023 the original author or authors.
+ *  Copyright 2023-2024 the original author or authors.
  *
  *  Licensed to the Apache Software Foundation (ASF) under one or more
  *  contributor license agreements. See the NOTICE file distributed with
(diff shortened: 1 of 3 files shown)


core/citrus-api/src/main/java/org/citrusframework/spi/Resources.java
```diff
@@ -263,11 +268,27 @@ public String getLocation() {
 
         @Override
         public boolean exists() {
+
+
```
bbortt, line 272:
```suggestion
```

please do also update the copyright year to `2023-2024` on top of this class.

core/citrus-api/src/test/java/org/citrusframework/spi/ResourcesTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.citrusframework.exceptions.CitrusRuntimeException;
+import org.citrusframework.spi.Resources.ByteArrayResource;
+import org.citrusframework.spi.Resources.ClasspathResource;
+import org.citrusframework.spi.Resources.FileSystemResource;
+import org.citrusframework.spi.Resources.UrlResource;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class ResourcesTest {
```
bbortt, line 31: this file is not properly formatted.

bbortt (changes requested):
thank you very much! looks good to me in general, only some nitpicks.

tschlat (commented):
Done

bbortt:
> Done

well thank you!

bbortt (approved):

christophd (approved):
Awesome!

PR #1117, opened 2024-01-31
fix(#1115): Create new test context for each Cucumber scenario

### LINKED ISSUE #1115: Citrus Cucumber Spring uses the same TestContext for all tests
**Citrus Version**
4.1.1

**Expected behavior**
Do not use the same TestContext for all Cucumber feature scenarios but create a fresh context for each scenario.

**Actual behavior**
The Cucumber Spring object factory provided by Citrus uses a singleton TestContext for all scenarios executed. This leads to unexpected behavior as scenarios influence each other when sharing the same context (e.g. when reporting exceptions to the context)

christophd:
- Citrus Cucumber Spring object factory must create new test context instance for each scenario
- Using a singleton test context instance for all scenarios leads to unexpected behavior (e.g. when exceptions are added to the test context)

Fixes #1115


christophd:
Fixes #1115

Issue #1122, opened 2024-02-14
Test-Context Error handling is not graceful


bbortt:
**Citrus Version**

`main@0c48227`

**Expected behavior**

`TestContext#handleError(String, String, String, Exception)` returns the original exception, no matter what.

Link: https://github.com/citrusframework/citrus/blob/main/core/citrus-api/src/main/java/org/citrusframework/context/TestContext.java#L435.

**Actual behavior**

Whenever one of the listeners fails, this exception will be thrown instead of the original one. This `testListeners...` 3-tuple must be "graceful", catch any exceptions that would occur during invocation and handle it. Finally return the original `cause`.

PR #1123, opened 2024-02-14
Fix: Test-Context Error handling is not graceful #1122

repanareddysekhar:


core/citrus-api/src/main/java/org/citrusframework/context/TestContext.java
```diff
@@ -443,7 +443,7 @@ public CitrusRuntimeException handleError(String testName, String packageName, S
         testListeners.onTestFailure(dummyTest, exception);
         testListeners.onTestFinish(dummyTest);
 
-        return exception;
+        return new RuntimeException(cause);
```
bbortt, line 446: welcome, new contributor! before anything: thanks for your help - much appreciated!

however, that is not exactly what I meant within #1122 😉 the signature of the method was just fine. on the other hand, I would expect [this `CitrusRuntimeException`](https://github.com/citrusframework/citrus/pull/1123/files#diff-1b6d909722922f98386bb17d7115b1d483c9b18af54141597b064f4e71c03b99R439) to be returned, no matter what. but [here](https://github.com/citrusframework/citrus/pull/1123/files#diff-1b6d909722922f98386bb17d7115b1d483c9b18af54141597b064f4e71c03b99R441) come the problematic lines:

```java
// inform test listeners with failed test
testListeners.onTestStart(dummyTest);
testListeners.onTestFailure(dummyTest, exception);
testListeners.onTestFinish(dummyTest);
```

if you want to take on that issue, I would suggest starting with the following test in `TestContextTest` (note that it fails right now):

```java
@Test
public void handleErrorDespiteListenerFailure() {
    var testListeners = new TestListeners();

    var testStartListenerMock = mock(TestListener.class);
    testListeners.addTestListener(testStartListenerMock);

    context.setTestListeners(testListeners);

    var cause = new Exception("thrown with a purpose!");
    doThrow(cause).when(testStartListenerMock).onTestStart(any(TestCase.class));

    var testName = "test name";
    var packageName = "package name";

    var message = "additional message";

    var citrusRuntimeException = Assert.expectThrows(CitrusRuntimeException.class, () -> context.handleError(testName, packageName, message, cause));
    Assert.assertEquals(citrusRuntimeException.getMessage(), message);
    Assert.assertEquals(citrusRuntimeException.getCause(), cause);
}
```

repanareddysekhar: The above test is failing.
If we remove 
doThrow(cause).when(testStartListenerMock).onTestStart(any(TestCase.class));
this line it is passing.

I am missing something here about requirement @bbortt

bbortt: yes, but how would you change the code, so that it does not? the invocations of `testListeners#...` should be handled gracefully.

PR #1131, opened 2024-03-06
fix: parameter list parsing

novarx:


core/citrus-base/src/test/java/org/citrusframework/functions/FunctionUtilsTest.java
```diff
@@ -16,84 +16,111 @@
...
+        assertEquals(resolveFunction("citrus:upperCase('Yes', 'I like Citrus!')", context), "YES");
+        assertEquals(resolveFunction("citrus:concat('Hello Yes, I like Citrus!')", context), "Hello Yes, I like Citrus!");
+        assertEquals(resolveFunction("citrus:concat('Hello Yes,I like Citrus!')", context), "Hello Yes,I like Citrus!");
+        assertEquals(resolveFunction("citrus:concat('Hello Yes,I like Citrus, and this is great!')", context), "Hello Yes,I like Citrus, and this is great!");
+        assertEquals(resolveFunction("citrus:concat('Hello Yes , I like Citrus!')", context), "Hello Yes , I like Citrus!");
+        assertEquals(resolveFunction("citrus:concat('Hello Yes, I like Citrus!', 'Hello Yes,we like Citrus!')", context), "Hello Yes, I like Citrus!Hello Yes,we like Citrus!");
+        assertEquals(resolveFunction("citrus:concat('Hello Yes, I like Citrus, and this is great!', 'Hello Yes,we like Citrus, and this is great!')", context), "Hello Yes, I like Citrus, and this is great!Hello Yes,we like Citrus, and this is great!");
+
+//        assertEquals(resolveFunction("citrus:upperCase(''Yes, I like Citrus!)", context), "''YES, I LIKE CITRUS!");
```
novarx, line 87: I don't really understand those three test - are they really intended like that?

Because compared i.E. to CSV (if I recall correct), if quotes are present in a string, they need to be inside quotes.

core/citrus-api/src/main/java/org/citrusframework/functions/FunctionParameterHelper.java
```diff
@@ -34,62 +33,94 @@ private FunctionParameterHelper() {}
...
     /**
      * Convert a parameter string to a list of parameters.
-     * 
+     *
      * @param parameterString comma separated parameter string.
      * @return list of parameters.
      */
     public static List<String> getParameterList(String parameterString) {
-        List<String> parameterList = new ArrayList<>();
+        return new ParameterParser(parameterString).parse();
```
novarx, line 41: The longer I think about that... wouldn't it be easier to just use a CSV-Parser for that?

novarx: Hey @bbortt 
Your thoughts on that?

core/citrus-base/src/test/java/org/citrusframework/functions/FunctionUtilsTest.java
bbortt: so, apparently unequal amount of quotes were accepted / ignored. and these few tests here do no longer pass, because of the provided fix in the parser. is that what you mean?

core/citrus-api/src/main/java/org/citrusframework/functions/FunctionParameterHelper.java
bbortt: hm.. I think CSV uses double quotes for concatenation. that's a bit problematic in the context of Java. other than that, I do too think that it's the same parsing logic.

PR #1135, opened 2024-03-17
gracefully shutdown test context

### LINKED ISSUE #1122: Test-Context Error handling is not graceful

bbortt:
closes #1122.

### DIFF
--- a/core/citrus-api/src/test/java/org/citrusframework/context/TestContextUnitTest.java
+++ b/core/citrus-api/src/test/java/org/citrusframework/context/TestContextUnitTest.java
@@ -1,0 +1,87 @@
+package org.citrusframework.context;
+
+import org.citrusframework.exceptions.CitrusRuntimeException;
+import org.citrusframework.report.TestListener;
+import org.citrusframework.report.TestListeners;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doThrow;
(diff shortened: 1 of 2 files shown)


core/citrus-api/src/test/java/org/citrusframework/context/TestContextUnitTest.java
```diff
@@ -0,0 +1,87 @@
...
+        doThrow(cause).when(testListenerMock).onTestFinish(any(TestContext.EmptyTestCase.class));
+
+        invokeHandleErrorOnFixture(cause);
+
+        verify(testListenerMock).onTestStart(any(TestContext.EmptyTestCase.class));
+        verify(testListenerMock).onTestFailure(any(TestContext.EmptyTestCase.class), any(CitrusRuntimeException.class));
+        verify(testListenerMock).onTestFinish(any(TestContext.EmptyTestCase.class));
+    }
+
+    private TestListener fixtureWithTestListeners() {
```
tschlat, line 66: Just to add something to this review:
the name of this method implies, that it returns the fixture wired with listeners. Instead it does exactly this, but returns the listeners.

So maybe use: attachTestListenerMockToFixture

PR #1136, opened 2024-03-20
chore: Update Kubernetes module

christophd:
- Update Kubernetes Fabric8 client to 6.10.0


bbortt:
I think the copyrights on the touched classes should be updated. and if you don't mind me saying this, writing some tests would do good 😉 I can spend some if you give me time over the weekend.

christophd:
@bbortt thanks! about the copyright year. How about removing the year information completely? it does not give too much information because you can always have a look into the GitHub history when a file has been added

bbortt:
ahm yea. but isn't it necessary to comply with the license? that each file must mention the licensing? I am no lawyer, but that's how I always understood it.

christophd:
Yeah right, I did not mean to remove the license header. just the copyright year information that needs to be adjusted every time

christophd:
I have enabled the integration tests now with using the Fabric8 mock K8s server

PR #1142, opened 2024-04-09
reduce logging level of successful validations

bbortt:
I found the validation log levels to be rather annoying. I only want to know about this if something goes wrong, not if something passes. that should imo be something for a report, but not for a log.


christophd (commented):
yes, I agree that INFO level is not appropriate for this kind of log output. Why not using DEBUG level? What I dislike about TRACE level is that it is barely enabled because other libraries then flush a ton of output and you can barely analyze and review the logs.

I would vote for using DEBUG level here.

bbortt:
well, you can always enable `TRACE` for certain packages only. but I am without strong opinion on this: I will update it to `DEBUG`.

Issue #1143, opened 2024-04-10
Ensuring Unique Message IDs in Citrus Templates for Reliable Test Correlation


tschlat:
**Citrus Version**
4.x

**Expected behavior**

Given:
An XML test case calling a template with a send message action.

When:
Utilizing this template across various XML tests.

Then:
The sent messages should exhibit unique message IDs.

The rationale behind this expectation is crucial. The message ID assigned to each sent message serves as a means to correlate it with the corresponding received message, facilitated by the correlation manager. In scenarios where tests execute concurrently and messages share identical IDs, inbound messages might erroneously correlate with incorrect outgoing messages, potentially leading to test failures.

** Actual Behavior **
Currently, messages sent from templates consistently possess identical message IDs. This behavior stems from the HttpSendRequestActionParser, which generates a BeanDefinition featuring a fixed HttpMessage. Despite the builder consistently crafting new SendMessageAction instances, the message creation relies on the original HttpMessage residing within the BeanDefinition. Consequently, the IDs of messages generated from the SendMessageAction consistently match, leading to the observed behavior.

PR #1144, opened 2024-04-11
fix(#1143): let HttpMessageBuilder create messages with unique ids

tschlat:
I've updated the HttpMessageBuilder to consistently generate messages with new IDs and timestamps. As a result, messages created by the builder will always have unique identifiers and may differ in their timestamps. I haven't identified any scenarios where this change would cause any issues.

I would appreciate Christoph's thorough review and expert opinion on this matter, considering the potential for unforeseen side effects.

### DIFF
--- a/endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
+++ b/endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
@@ -73,7 +73,17 @@ public class HttpMessage extends DefaultMessage {
      * @param message The base message for the copy operation
      */
     public HttpMessage(final Message message) {
-        super(message);
+        this(message, false);
+    }
+
+    /**
+     * Constructs a new HTTP message based on the provided Message object, with an option to force Citrus header update.
+     *
+     * @param message the Message object to copy
+     * @param forceCitrusHeaderUpdate flag indicating whether to force Citrus header update
+     */
+    public HttpMessage(final Message message, boolean forceCitrusHeaderUpdate) {
+        super(message, forceCitrusHeaderUpdate);
         copyCookies(message);
     }
 
@@ -102,8 +112,8 @@ public class HttpMessage extends DefaultMessage {
      * @param message the message to extract the cookies from
      */
     private void copyCookies(final Message message) {
-        if (message instanceof HttpMessage) {
-            this.cookies.putAll(((HttpMessage) message).getCookiesMap());
+        if (message instanceof HttpMessage httpMessage) {
+            this.cookies.putAll(httpMessage.getCookiesMap());
         }
     }
 
@@ -137,8 +147,9 @@ public class HttpMessage extends DefaultMessage {
      */
     public HttpMessage status(final HttpStatusCode statusCode) {
         setHeader(HttpMessageHeaders.HTTP_STATUS_CODE, statusCode.value());
-        if (HttpStatus.resolve(statusCode.value()) != null) {
-            setHeader(HttpMessageHeaders.HTTP_REASON_PHRASE, HttpStatus.resolve(statusCode.value()).name());
+        HttpStatus status = HttpStatus.resolve(statusCode.value());
+        if (status != null) {
+            setHeader(HttpMessageHeaders.HTTP_REASON_PHRASE, status.name());
         }
         return this;
     }
@@ -503,9 +514,12 @@ public class HttpMessage extends DefaultMessage {
     /**
      * Reads request from complete request dump.
      *
+     * <p>Note that this method is called from YAKS.
+     *
      * @param requestData The request dump to parse
      * @return The parsed dump as HttpMessage
      */
+    @SuppressWarnings("unused")
     public static HttpMessage fromRequestData(final String requestData) {
         try (final BufferedReader reader = new BufferedReader(new StringReader(requestData))) {
             final HttpMessage request = new HttpMessage();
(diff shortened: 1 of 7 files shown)


endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
```diff
@@ -555,10 +566,7 @@ public static HttpMessage fromResponseData(final String responseData) {
     }
 
     private void addQueryParam(final String name, final String value) {
-        if (!this.queryParams.containsKey(name)) {
-            this.queryParams.put(name, new LinkedList<>());
-        }
-        this.queryParams.get(name).add(value);
+        this.queryParams.computeIfAbsent(name, k->new LinkedList<>()).add(value);
```
bbortt, line 569:
```suggestion
        this.queryParams.computeIfAbsent(name, k -> new LinkedList<>()).add(value);
```

endpoints/citrus-http/src/test/java/org/citrusframework/http/message/HttpMessageBuilderTest.java
```diff
@@ -17,29 +17,31 @@
...
-import static org.testng.Assert.assertTrue;
-
 public class HttpMessageBuilderTest {
 
     private HttpMessage message;
 
     @BeforeMethod
     public void setUp(){
         message = new HttpMessage("");
+        message.setHeader(MessageHeaders.TIMESTAMP, System.currentTimeMillis()-1);
```
bbortt, line 44:
```suggestion
        message.setHeader(MessageHeaders.TIMESTAMP, System.currentTimeMillis() - 1);
```

endpoints/citrus-http/src/test/java/org/citrusframework/http/message/HttpMessageBuilderTest.java
```diff
@@ -53,8 +55,10 @@ public void testDefaultMessageHeader() {
 
         //THEN
         assertEquals(builtMessage.getHeaders().entrySet().size(), 3);
-        assertEquals(message.getHeader(MessageHeaders.ID), builtMessage.getHeader(MessageHeaders.ID));
-        assertEquals(message.getHeader(MessageHeaders.TIMESTAMP), builtMessage.getHeader(MessageHeaders.TIMESTAMP));
+        assertNotNull(message.getHeader(MessageHeaders.ID));
+        assertNotNull(message.getHeader(MessageHeaders.TIMESTAMP));
+        assertNotEquals(message.getHeader(MessageHeaders.ID), builtMessage.getHeader(MessageHeaders.ID));
+        assertNotEquals(message.getHeader(MessageHeaders.TIMESTAMP), builtMessage.getHeader(MessageHeaders.TIMESTAMP));
```
bbortt, line 61: I think the negative case (`forceCitrusHeaderUpdate` being `false`) deserves a test too 😉

bbortt (changes requested):
as discussed, I think this change adequate. would be interesting to hear @christophd opinion, since he's clearly better with historical stuff than I am 😛 

@tschlat and some nit-picks as usual.

endpoints/citrus-http/src/test/java/org/citrusframework/http/message/HttpMessageBuilderTest.java
tschlat: I added respective test cases at DefaultMessage and HttpMessage level.

endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
```diff
@@ -77,6 +77,17 @@ public HttpMessage(final Message message) {
         copyCookies(message);
```
bbortt, line 77: this constructor could just contain `this(message, false)`, right? just to have said it.

bbortt (approved):
/lgtm

christophd (commented):
I do not remember exactly why this has been implemented to preserve the message ID and TIMESTAMP in HttpMessageBuilder. I just remember that the static message builder may be used several times in a kind of hierarchy and each level would then create its own IDs. I can imagine that this may cause issues when the user extracts the ID saving it to a test variable and referencing it at a later time in the test. But I can not point you exactly to this where this is being used.

In general I would rely on the unit and integration tests that we have. If those tests are green I am good to proceed with the change.

Maybe we can expose of the default for `forceCitrusHeaderUpdate` into a new setting that can be adjusted via application.properties and env variables. Then we should be good to also restore the old behavior when a user needs to do so.

WDYT?

endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
tschlat: That is - indeed true. I adjusted it accordingly.

endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
```diff
@@ -534,7 +548,9 @@ public static HttpMessage fromRequestData(final String requestData) {
      *
      * @param responseData The response dump to parse
      * @return The parsed dump as HttpMessage
+     * @deprecated this method is not used and will be removed in the next major release
      */
+    @Deprecated(forRemoval = true) // TODO: @Christoph is it ok to mark this as deprecated? Doesn't seem to be in use.
```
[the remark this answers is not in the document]
tschlat: same here

endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpMessage.java
```diff
@@ -505,7 +517,9 @@ public HttpMessage cookie(final Cookie cookie) {
      *
      * @param requestData The request dump to parse
      * @return The parsed dump as HttpMessage
+     * @deprecated this method is not used and will be removed in the next major release
      */
+    @Deprecated(forRemoval = true) // TODO: @Christoph is it ok to mark this as deprecated? Doesn't seem to be in use.
```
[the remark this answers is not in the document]
tschlat: The IDE told me, that the method is not used and after double checking this seemed true. So this was meant as a cleanup. I removed the deprecation and added a suppression plus a remark in the java doc.

bbortt:
@tschlat thanks, I've squashed your commits.

PR #1153, opened 2024-04-25
pipeline updates

bbortt:
nightly dependency reports, as superset of https://github.com/citrusframework/citrus/pull/901.


.github/workflows/lts.yml
```diff
@@ -19,10 +19,10 @@ name: lts_preview
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * 1"
+    - cron: "0 0 * * *"
```
christophd, line 22: The lts_preview workflow is not fully stable. does it make sense to run this every day? We get emails for failing jobs then every day

christophd (approved):

.github/workflows/lts.yml
bbortt: kk, I am fine with that.

PR #1154, opened 2024-04-25
json message validator access

bbortt:
we have a custom JSON validator and want to construct the error message in the same format as the original: https://github.com/citrusframework/citrus/pull/1154/files#diff-f9be5af8b22ec2ff3b7b516d047768e9d3bdbfaea29748310aaa76ff125b6600R89.


christophd (approved):
Many people still use Spring XML bean definitions that use the public setters when adding `property` settings. I have no strong opinion but this would break peoples XML bean config then in case we change the setters to private visibility

bbortt:
> Many people still use Spring XML bean definitions that use the public setters when adding property settings. I have no strong opinion but this would break peoples XML bean config then in case we change the setters to private visibility

good point!

PR #1155, opened 2024-05-01
fix: validate json array order if strict mode

novarx:
See #1157
A follow up to https://github.com/citrusframework/citrus/pull/1102
It seems that not validating the order of json arrays if `strict` is not the desired behaviour.

see https://github.com/citrusframework/citrus/pull/1102#discussion_r1458473481

### DIFF
--- a/src/manual/validation-json.adoc
+++ b/src/manual/validation-json.adoc
@@ -1,16 +1,16 @@
 [[json-message-validation]]
 == Json validation
 
-Message formats such as Json have become very popular, in particular when dealing with RESTful services. Citrus is able to
-expect and validate Json messages with a powerful comparison of Json structures.
+Message formats such as Json have become very popular, in particular when dealing with RESTful services.
+Citrus is able to expect and validate Json messages with a powerful comparison of Json structures.
 
-IMPORTANT: By default, Citrus will use XML message formats when sending and receiving messages. This also reflects to the
-message validation logic Citrus uses for incoming messages. So by default Citrus will try to parse the incoming message as
-XML DOM element tree. In case we would like to enable Json message validation we have to tell Citrus that we expect a Json
-message right now.
+IMPORTANT: By default, Citrus will use XML message formats when sending and receiving messages.
+This also reflects to the message validation logic Citrus uses for incoming messages.
+So by default Citrus will try to parse the incoming message as XML DOM element tree.
+In case we would like to enable Json message validation we have to tell Citrus that we expect a Json message right now.
 
-Json message validation is not enabled by default in your project. You need to add the validation module to your project
-as a Maven dependency.
+Json message validation is not enabled by default in your project.
+You need to add the validation module to your project as a Maven dependency.
 
 .Json validation module dependency
 [source,xml]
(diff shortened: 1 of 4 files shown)


novarx:
@bbortt

etieskrill:
There is the need to configure the json array ordering separately from the `strict` mode. This should be configurable on a per-test basis. Probably by adding a new field to `org.citrusframework.validation.json.JsonMessageValidationContext`, something like `checkArrayOrder`, and choosing the validation method for array items based off of the new property, instead of `strict`. If the property was not set explicitly, the default should be `false` if non-strict, and `true` if strict, i think.

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
```diff
@@ -110,7 +110,10 @@ private void validateJSONArray(JsonElementValidator validator, JsonElementValida
             validateSameSize(control.getJsonPath(), arrayControl.expected, arrayControl.actual);
         }
         for (int i = 0; i < arrayControl.expected.size(); i++) {
-            if (!isAnyValidItemInActualArray(validator, arrayControl, i)) {
+            var isValid = strict
+                    ? isValidItem(new JsonElementValidatorItem<>(i, arrayControl.actual.get(i), arrayControl.expected.get(i)).parent(control), validator)
+                    : isAnyValidItemInActualArray(validator, arrayControl, i);
```
tschlat, line 115: In general I'd say `isAnyValidItemInActualArray` is not sufficient to determine correct order of an array in non strict mode.  

This should be revrealed by:

```
control = [1, 1 ,2, 1]
actual= [1, 2, 2, 2]
```

To be non breaking, you should implement the behaviour as it was before. Strict was only used to verify the number of elements in the array. So if the received array had trailing elements, this was OK in non strict mode. But the order was still verified for all elements of the control array. If we change this behaviour as proposed, we are risking false positives. Also, in general, this would not support our use case. We need to be able to individually decide whether we want to neglect sorting. Strict/non strict works on a global level.

bbortt: > It's essential to remember that strict mode should not dictate how the order of an array is validated.

don't you think this strange? 👯 that makes no sense. strict is strict. period.

tschlat: I don't know what you are refering to to. Strict was in the past only used to tell the validator whether validation should fail on additional elements. This holds true for objects and also for arrays. 

To makt this clear: up to my knowledge there has never been a way to tell the validator to ignore ordering of arrays.

novarx: Array Order now always gets validated (as before #1102).

And mentioned Testcase was added

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
```diff
@@ -96,21 +96,28 @@ static boolean isIgnoredByPlaceholderOrExpressionList(Collection<String> ignoreE
...
         }
+        int actualIndex = 0;
         for (int i = 0; i < arrayControl.expected.size(); i++) {
-            if (!isAnyValidItemInActualArray(validator, arrayControl, i)) {
+            if (isIgnoredByPlaceholderOrExpressionList(ignoreExpressions, arrayControl.child(i, i))) {
+                continue;
+            }
+            boolean isValid = false;
+
+            while (!isValid && actualIndex < arrayControl.actual.size()) {
```
tschlat, line 114: Without judging whether this approach would make more sense, it is important to notice, that it is different to the previous one. So it is potentially breaking and may result in false positives for tests which are out there. Since there is no need for this "potential improvement", I would simply stay with the previous implementation which only accepted "additional elements" at the end of the array in non strict mode. If you really change the behaviour like this, you need to properly document it and provide a configuration option to switch back to the previous behaviour.

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
```diff
@@ -72,6 +71,14 @@ public static JsonAssertion[] validJsonPairsIfNotStrict() {
                 new JsonAssertion(
```
tschlat, line 71: I am pretty sure, that this would have failed in the past.

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
```diff
@@ -72,6 +71,14 @@ public static JsonAssertion[] validJsonPairsIfNotStrict() {
...
                         "[1, 2, 3]",
                         "[2] "
+                ),
+                new JsonAssertion(
```
tschlat, line 75: This as well.

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
bbortt: because it did only take suffixes into account, but prefixes were "strict"?

novarx: @bbortt yes exactly - I'll do my best to document that a bit more concise.

validation/citrus-validation-json/src/main/java/org/citrusframework/validation/json/JsonElementValidator.java
novarx: As discussed bilateral, we'll keep the "new" behaviour and document it.

src/manual/validation-json.adoc
```diff
@@ -25,23 +25,25 @@ as a Maven dependency.
...
+- The order of Json object properties can differ, whereas array order is always validated.
+- Test variables and ignored-placeholders (`@ignore@`) can be used.
+- JsonArray as well as nested JsonObjects are supported.
+GroovyJsonMessageValidator:: Extended groovy message validator provides specific Json slurper support.
+With Json slurper the tester can validate the Json message body with closures for instance.
+
+IMPORTANT: The Json validator offers two different modes to operate.
+By default, *strict* mode is enabled and the validator will also check the exact amount of entries (object properties / array items) to match in received and control message.
+No additional fields in received Json data structure will be accepted.
+In *soft* mode the validator allows additional entries in received Json, so the control Json can be a subset of the recieved.
```
tschlat, line 38:
```suggestion
In *soft* mode the validator allows additional entries (object properties / array items) in received Json, so the control Json can be a subset of the recieved.
```

tschlat (changes requested):
lgtm

To have this properly documented, I kindly ask to create an issue to document the change, that non strict json validation now allows additional elements at any position in the array, while still enforcing the correct order of the control array elements.

bbortt (approved):
I'm fine with the current solution, knowing that @etieskrill is working on another feature to make this even more customizable.

PR #1167, opened 2024-05-29
Fix: Use ISO date format for random JSON property values (OpenAPI)

### LINKED ISSUE #1166: OpenAPI payload in request body has wrong date format
**Citrus Version**
4.2.1

**Expected behavior**
When sending a request using the OpenApiRequestActionBuilder, random data is put into the JSON. When the format is `date`, the format should be `yyyy-MM-dd`.   

**Actual behavior**
The date format is `dd-MM-yyyy`, which leads to deserialization issues at the HTTP server.

**Test case sample**
```java
t.when(
        openapi()
                        .specification(apiSpecification)
                        .client(openApiClient)
                        .send("createTodo")
);

```

ueberfuhr:
The `OpenApiTestDataGenerator` creates random values for JOSN properties. When a property has format `date`, the function `citrus:currentDate()` was used, which is wrong, because it leads to an invalid date pattern. I fixed this by specifying the correct pattern.

Resolves #1166

### DIFF
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiTestDataGenerator.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiTestDataGenerator.java
@@ -32,6 +32,7 @@ import org.springframework.util.StringUtils;
  * enforce the specification rules.
  *
  * @author Christoph Deppisch
+ * @author Ralf Ueberfuhr
  */
 public class OpenApiTestDataGenerator {
 
@@ -119,7 +120,7 @@ public class OpenApiTestDataGenerator {
             }
 
             if (schema.format != null && schema.format.equals("date")) {
-                payload.append("citrus:currentDate()");
+                payload.append("citrus:currentDate('yyyy-MM-dd')");
             } else if (schema.format != null && schema.format.equals("date-time")) {
                 payload.append("citrus:currentDate('yyyy-MM-dd'T'hh:mm:ss')");
             } else if (StringUtils.hasText(schema.pattern)) {


connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiTestDataGenerator.java
```diff
@@ -119,7 +120,7 @@ public static String createRandomValueExpression(OasSchema schema, Map<String, O
             }
 
             if (schema.format != null && schema.format.equals("date")) {
-                payload.append("citrus:currentDate()");
+                payload.append("citrus:currentDate('yyyy-MM-dd')");
```
bbortt, line 123: for reference:

> date – full-date notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, 2017-07-21

[source](https://swagger.io/docs/specification/data-models/data-types/#string).

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiTestDataGenerator.java
```diff
@@ -32,6 +32,7 @@
  * enforce the specification rules.
  *
  * @author Christoph Deppisch
+ * @author Ralf Ueberfuhr
```
bbortt, line 35:
```suggestion
```

sorry to bother you again, but please remove these. see https://github.com/citrusframework/citrus/pull/1171#discussion_r1619210698.

bbortt (changes requested):

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiTestDataGenerator.java
ueberfuhr: Okay, if this is disturbing, I'll remove the @author tag (but only my own) to minimize the diffs.

PR #1169, opened 2024-05-29
Fix: use explicitly declared values instead of random values (OpenAPI)

### LINKED ISSUE #1168: OpenAPI request always generates random data, even if a body is specified
**Citrus Version**
4.2.1

**Expected behavior**
When I specify a body for the request, this should be sent to the server, independent from being valid in comparison to the OpenAPI schema.

**Actual behavior**
Random data is sent to the server.

**Test case sample**
```java
        t.when(
                openapi()
                        .specification(apiSpecification)
                        .client(openApiClient)
                        .send("createTodo") // operationId
                        .message()
                        .body("{\"title\": \"test\"}")
        );
```

ueberfuhr:
When generating the request message, random data is used for the payload and for parameters - even when values are explicitly specified. I changed this behavior to prefer explicitly declared values and to skip random data generation in this case.

Resolves #1168

### DIFF
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/actions/OpenApiClientRequestActionBuilder.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/actions/OpenApiClientRequestActionBuilder.java
@@ -121,7 +121,7 @@ public class OpenApiClientRequestActionBuilder extends HttpClientRequestActionBu
                         });
             }
 
-            if(httpMessage.getPayload() == null) {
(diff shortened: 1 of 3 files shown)


bbortt:
I think that a test would do good here too.. but honestly, I have no clue if something exists at all. did you check @ueberfuhr?

ueberfuhr:
@bbortt - I have added a test. Please review.

bbortt (approved):
thanks for the test. that helps a lot. amazing contribution @ueberfuhr ❤️

Issue #1175, opened 2024-06-11
OpenApiActionBuilder enhancements


tschlat:
**User story**
To support citrus simulator in generating random messages for OpenApi, several adjustments to the OpenApi Connector need to be implemented.

Such as:
- an OpenApiRepository for storage of OpenApis
- adjustments to OpenApi action builders to be able to properly guess a random response to generate
- Validation of Request/Response, using SchemaValidator

PR #1179, opened 2024-06-21
Adds new changes to main #1175 Branch: OpenApiRepository and Validation

tschlat:
Implement new features:
- a repository for OpenApis
- validation based on atlassian OpenAPI validator framework

Note that with respect to validation, there are now two different validators. One which is based on standard citrus validation mechanism and one based on atlassian framework. The later is much more powerfull but might raise more issues. We should make a decision on how to proceed here:

- Remove standard citrus implementation and rely on atlassian validation
- Use both as is. Since atlassian validation is performed message level, it comes first and citrus most probably is never called.
- Have cirtus kick in, when atlassian is disabled (possible by configuration and on a per request basis)

### DIFF
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiSpecificationAdapter.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiSpecificationAdapter.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.openapi;
+
+/**
+ * Adapter class that links an OAS entity to its associated OpenAPI specification context.
+ * This class provides methods to access both the OpenAPI specification and the specific OAS entity.
+ *
+ * @param <T> the type to which the specification is adapted.
+ */
+public class OpenApiSpecificationAdapter<T> {
+    
+    private final OpenApiSpecification openApiSpecification;
+
+    private final T entity;
+
+    public OpenApiSpecificationAdapter(OpenApiSpecification openApiSpecification, T entity) {
+        this.openApiSpecification = openApiSpecification;
+        this.entity = entity;
+    }
+
+    public OpenApiSpecification getOpenApiSpecification() {
+        return openApiSpecification;
+    }
+
+    public T getEntity() {
+        return entity;
+    }
+}
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiSpecificationAdapter.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiSpecificationAdapter.java
@@ -22,22 +22,6 @@ package org.citrusframework.openapi;
  *
  * @param <T> the type to which the specification is adapted.
  */
-public class OpenApiSpecificationAdapter<T> {
-    
-    private final OpenApiSpecification openApiSpecification;
+public record OpenApiSpecificationAdapter<T>(OpenApiSpecification openApiSpecification, T entity) {
 
-    private final T entity;
-
-    public OpenApiSpecificationAdapter(OpenApiSpecification openApiSpecification, T entity) {
-        this.openApiSpecification = openApiSpecification;
-        this.entity = entity;
-    }
-
-    public OpenApiSpecification getOpenApiSpecification() {
-        return openApiSpecification;
-    }
-
-    public T getEntity() {
-        return entity;
-    }
 }
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/actions/OpenApiClientRequestActionBuilder.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/actions/OpenApiClientRequestActionBuilder.java
@@ -16,17 +16,13 @@
 
 package org.citrusframework.openapi.actions;
 
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Optional;
-import java.util.regex.Pattern;
-
-import io.apicurio.datamodels.openapi.models.OasDocument;
 import io.apicurio.datamodels.openapi.models.OasOperation;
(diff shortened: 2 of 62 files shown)


bbortt:
> We should make a decision on how to proceed here:

I am personally ok with "hardening" features when it comes to **feature releases** - which it is.

> The later is much more powerfull but might raise more issues.

I don't think that has to be breaking necessarily, as long as the API itself (e.g. OpenAPI must be stored in a different directory, leading to false-positives because no spec is loaded at all) does not break.

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/actions/OpenApiClientRequestActionBuilder.java
```diff
@@ -56,6 +56,16 @@ public OpenApiClientRequestActionBuilder(OpenApiSpecification openApiSpec, Strin
     public OpenApiClientRequestActionBuilder(HttpMessage httpMessage, OpenApiSpecification openApiSpec,
                                              String operationId) {
         super(new OpenApiClientRequestMessageBuilder(httpMessage, openApiSpec, operationId), httpMessage);
+
+        openApiRequestValidationProcessor = new OpenApiRequestValidationProcessor(openApiSpec, operationId);
+        process(openApiRequestValidationProcessor);
+    }
+
+    public OpenApiClientRequestActionBuilder disableOasValidation(boolean b) {
```
bbortt, line 64: since this method is called "disable" I would just assume it disables validation (no matter what). I would not expect a parameter for sure. thoughts?

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/model/OasAdapter.java
```diff
@@ -0,0 +1,24 @@
+package org.citrusframework.openapi.model;
+
+import io.apicurio.datamodels.core.models.Node;
+
+public class OasAdapter<S extends Node, T> {
```
bbortt, line 5: this is a record as well. I've updated it:

```java
public record OasAdapter<S extends Node, T>(S node, T adapted) {

}
```

or do you intend to overwrite it somewhere?

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/model/v2/Oas20ModelHelper.java
```diff
@@ -157,6 +127,10 @@ public static Map<String, OasSchema> getHeaders(Oas20Response response) {
                 .collect(Collectors.toMap(OasHeader::getName, Oas20ModelHelper::getHeaderSchema));
     }
 
+    private static boolean isFormDataMediaType(String type) {
+        return Arrays.asList("application/x-www-form-urlencoded", "multipart/form-data").contains(type);
```
bbortt, line 131: out of curiosity, where did you get these from? because there exist more, I think. at least some `message/*` things.

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/validation/OpenApiRequestValidator.java
```diff
@@ -0,0 +1,97 @@
...
+
+        SimpleRequest.Builder finalRequestBuilder = requestBuilder;
+        finalRequestBuilder.withAccept(httpMessage.getAccept());
+
+        httpMessage.getQueryParams()
+            .forEach((key, value) -> finalRequestBuilder.withQueryParam(key, new ArrayList<>(
+                value)));
+
+        httpMessage.getHeaders().forEach((key, value) -> {
+            if (value instanceof Collection<?>) {
```
bbortt, line 86: I've changed this to:

```suggestion
            if (value instanceof Collection<?> collection) {
```

connectors/citrus-openapi/src/test/java/org/citrusframework/openapi/OpenApiTestDataGeneratorTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+// TODO: Add more tests
+public class OpenApiTestDataGeneratorTest {
+
+    private final OpenApiSpecification pingSpec = OpenApiSpecification.from(
+        Resources.create("classpath:org/citrusframework/openapi/ping/ping-api.yaml"));
+
+    // TODO: fix this by introducing mature validation
```
bbortt, line 36: what about these? are these to be based on the decision?

core/citrus-base/src/main/java/org/citrusframework/util/StringUtils.java
```diff
@@ -43,7 +45,7 @@ public static boolean isEmpty(String str) {
         return str == null || str.isEmpty();
     }
 
-    public static String appendSegmentToPath(String path, String segment) {
+    public static String appendSegmentToUrlPath(String path, String segment) {
```
bbortt, line 48: should be non breaking, I think:

```java
    @Deprecated(forRemoval = true)
    public static String appendSegmentToPath(String path, String segment) {
        appendSegmentToUrlPath(path, segment);
    }

    public static String appendSegmentToUrlPath(String path, String segment) {
```

bbortt (changes requested):
not an easy one! well done @tschlat. especially the reported coverage on `citrus-openapi`:

image

I took the freedom to commit some minor adjustments. nothing big, overall.

I've also questioned some code parts, but only one or two.

core/citrus-base/src/main/java/org/citrusframework/util/StringUtils.java
bbortt: (I have _not_ updated this)

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/actions/OpenApiServerActionBuilder.java
```diff
@@ -137,11 +151,11 @@ public TestActionBuilder<?> getDelegate() {
      */
     @Override
     public void setReferenceResolver(ReferenceResolver referenceResolver) {
-        if (referenceResolver == null) {
+        if (referenceResolver != null) {
```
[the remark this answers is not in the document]
bbortt: see https://github.com/citrusframework/citrus/pull/1184.

connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiSpecificationAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.citrusframework.openapi;
+
+/**
+ * Adapter class that links an OAS entity to its associated OpenAPI specification context.
+ * This class provides methods to access both the OpenAPI specification and the specific OAS entity.
+ *
+ * @param <T> the type to which the specification is adapted.
+ */
+public class OpenApiSpecificationAdapter<T> {
```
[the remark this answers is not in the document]
tschlat: No it's OK. I had this in mind but overcome it somehow.

bbortt (approved):

Issue #1181, opened 2024-06-24
Allow creation of TestResult objects with specific parameters


StenAtGH:
**User story**
As a back-end tester, I want to be able to pass additional parameters into the TestResult parameters, so that they can be integrated into reports subsequently. 

**Additional context**
TestResult already  has a parameter map but it is not possible to fill this from the standard test execution path in citrus. The aim is to maintain the current behavior as default, but with the possibility to modify it if necessary.

PR #1182, opened 2024-06-24
feat(#1181): adds TestResultInstanceProvider to support creation of TestResults with parameters

tschlat:


core/citrus-base/src/main/java/org/citrusframework/DefaultTestCase.java
```diff
@@ -51,6 +51,8 @@ public class DefaultTestCase extends AbstractActionContainer implements TestCase
 
     private static final Logger logger = LoggerFactory.getLogger(DefaultTestCase.class);
 
+    private static final TestResultInstanceProvider DEFAULT_TEST_RESULT_INSTANCE_PROVIDER = new DefaultTestResultInstanceProvider();
```
bbortt, line 54: am I missing something, or where is this class? have been [searching for it](https://github.com/search?q=repo%3Acitrusframework%2Fcitrus%20DefaultTestResultInstanceProvider&type=code).

bbortt (changes requested):

core/citrus-base/src/main/java/org/citrusframework/DefaultTestCase.java
tschlat: It is here:
https://github.com/citrusframework/citrus/pull/1182/files#diff-62108b235ff16c88ad3d6c58aeba60b76f84d55e98f54abe596be03401f471c1R547

bbortt: haha - I thought "why is this pipeline passing?!"

damn 🤖

bbortt (approved):

PR #1184, opened 2024-06-24
adjust resolver methods to ignore nullable values

bbortt:


tschlat (changes requested):
You missed some. E.g. OpenApiActionBuilder.

In general I would like to raise the question if this isn't worth an AbstractDelegatingActionBuilder. You could then remove a lot of code and a single test of the abstract class would be sufficient. There are 13 Classes that implement DelegatingTestActionBuilder. It seems that only two of them are not ReferenceResolverAware and I wonder if that is intentional or just forgotten.

bbortt:
I've updated the classes, but not touched the two mentioned that did not yet implement the interface. they seem to be a bit different, because they also use specific delegates.

PR #1206, opened 2024-08-23
Include unmarshaller in Validation Processor

leonschenk:
I came across this bug and supply a fix for it.

### DIFF
--- a/validation/citrus-validation-xml/src/main/java/org/citrusframework/validation/xml/XmlMarshallingValidationProcessor.java
+++ b/validation/citrus-validation-xml/src/main/java/org/citrusframework/validation/xml/XmlMarshallingValidationProcessor.java
@@ -151,7 +151,7 @@ public abstract class XmlMarshallingValidationProcessor<T> extends AbstractValid
                         "please add proper validation logic");
             }
 
-            return new XmlMarshallingValidationProcessor<>() {
+            return new XmlMarshallingValidationProcessor<>(unmarshaller) {
                 @Override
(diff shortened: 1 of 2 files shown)


validation/citrus-validation-xml/src/main/java/org/citrusframework/validation/xml/XmlMarshallingValidationProcessor.java
```diff
@@ -151,7 +151,7 @@ public XmlMarshallingValidationProcessor<T> build() {
                         "please add proper validation logic");
             }
 
-            return new XmlMarshallingValidationProcessor<>() {
+            return new XmlMarshallingValidationProcessor<>(unmarshaller) {
```
bbortt, line 154: could you please describe the bug and the solution to me? I don't get it just from looking at this, sorry 😢

bbortt (changes requested):

validation/citrus-validation-xml/src/main/java/org/citrusframework/validation/xml/XmlMarshallingValidationProcessor.java
christophd: I think I understand the issue. The issue is that the builder does not set the unmarshaller on the validation processor when building the instance. As a result the unmarshaller is always resolved at runtime via reference resolver lookup and setting a custom unmarshaller on the builder is not possible. Good fix! Thanks a lot

christophd (approved):
Good catch, many thanks!

PR #1209, opened 2024-08-30
fix(#1199): respect custom mailserver on startup

### LINKED ISSUE #1199: Citrus 4: MailServer only binds to loopback interface
**Citrus Version**
4.x

**Expected behavior**
Mailserver binds to all interfaces, as in Citrus 3.x, or host is configurable

**Actual behavior**
Mailserver binds to loopback interface only and the host is not configurable

**Test case sample**
SMTP interception on a host which has its mail port bound to a non loopback interface. Upgrade from Citrus 3.x to 4.x.

When we upgraded from Citrus 3 to Citrus 4, our mail tests did not work anymore. The problem was due to the default behaviour of subethasmtp vs. GreenMail. In subethasmtp, when no host is given, the mail server binds to all interfaces, whereas GreenMail will bind only to loopback per default.

As the Citrus `MailServerBuilder` does not have functions to set the host manually, and the `MailServer` initialises the GreenMail server with a hardcoded host value of `null`, the Citrus `MailServer` will only bind to the loopback interface and not receive Mails from another interfaces. This can be fixed by adding a host method to the `MailServerBuilder` and setting the host value instead of `null` for the GreenMail initialisation in the [MailServer startup() function](https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-mail/src/main/java/org/citrusframework/mail/server/MailServer.java#L103)

On a side note: it also seems that if you use the `smtp(GreenMail smtpServer)` method from the `MailServerBuilder`, the given GreenMail server is overwritten in the [MailServer startup() function](https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-mail/src/main/java/org/citrusframework/mail/server/MailServer.java#L103) as there is a missing null check before initialising a new GreenMail instance.

bbortt:
closes #1199.

@e17gpy I wen't for this implementation and will skip this part (as reported):

> As the Citrus MailServerBuilder does not have functions to set the host manually, and the MailServer initialises the GreenMail server with a hardcoded host value of null, the Citrus MailServer will only bind to the loopback interface and not receive Mails from another interfaces.

I think for security reasons this is ok. whoever want's to bind to a specific interface should create a custom `MailServer` and use that - with the provided fix this is now possible. does this fit your needs?

### DIFF
--- a/endpoints/citrus-mail/src/main/java/org/citrusframework/mail/server/MailServer.java
+++ b/endpoints/citrus-mail/src/main/java/org/citrusframework/mail/server/MailServer.java
@@ -16,14 +16,6 @@
 
 package org.citrusframework.mail.server;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Stack;
-import java.util.stream.Collectors;
-import javax.xml.transform.Source;
-
 import com.icegreen.greenmail.mail.MailAddress;
 import com.icegreen.greenmail.user.GreenMailUser;
 import com.icegreen.greenmail.user.UserException;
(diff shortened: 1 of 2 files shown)


e17gpy:
Hi, this should fit our needs. Thanks for implementing.

Issue #1186, opened 2024-07-02
Kotlin support


pablopaul:
**User story**
>As a developer, I want to use Kotlin for citrus tests so that I can use the same language as I use for my day to day work.

Is it planned to support Kotlin in the feature? Also as dialect in the docs?

bbortt:
can you give me some more details what would be needed to do so? isn't it (theoretically) already useable (because it's "just Java")?

I understand that it's probably not "hardened" for Kotlin, tho. but since I myself am not using it at the moment it'd be very hard to "come up with something" without requirements.

thanks!

hmmlopez:
Maybe he means just to have examples in the documentation with Kotlin examples? 
I’ve worked with Kotlin and Citrus together without any problems. But indeed needed to figure some stuff out.

bbortt:
> Also as dialect in the docs?

so we'd focus on that part? yea well, could be possible.. thoughts @christophd?

pablopaul:
Thanks for your thoughts

> can you give me some more details what would be needed to do so? isn't it (theoretically) already useable (because it's "just Java")?

In the meantime I came to the same conclusion and also our experience show full compatibility so far, so my question came from a pretty uninfirmed background.

Adding Kotlin examples would be superb. A quickfix could be just to mention general compatiblity somewhere in the docs, maybe this would even grow the user base for people who dont know that this is already possible.

For me it would be fine to close the issue, and I let this decision in your hands @bbortt

PR #1214, opened 2024-09-19
mention kotlin support in docs

bbortt:
closes https://github.com/citrusframework/citrus/issues/1186


bbortt:
@christophd build is hanging.. maybe force-merge?

christophd:
It is because the CI Job does not trigger for changes in documentation only. We can safely merge. Thanks

Issue #998, opened 2023-09-24
variable citrus junitreports directory


bbortt:
as reported in https://github.com/citrusframework/citrus/issues/983, the `target/citrus-reports/junitreports/TEST-TestSuite.xml` report path should be adaptable using a property.

**note:** `citrus.html.report.directory=build/citrus-reports` does move the `citrus-test-results.html` file.

hmmlopez:
I've noticed the following, if I set the `citrus.report.directory` property in the **citrus-application.properties** it solves my issue with the reports being generated in target directory when using gradle. I then don't need to specify the specific html/junit/etc location.
For example: `citrus.report.directory=build/citrus-reports`
Maybe it's just worth mentioning this in the documentation where gradle configuration is being explained?

bbortt:
that's a nice catch!

hmmlopez:
We can close this issue as well.

PR #1216, opened 2024-09-24
mention citrus reporting directory property

bbortt:
in addition to #998.


christophd:
Changes in documentation will not trigger the CI Job. We can safely merge. Thanks

bbortt:
ok, that's why. thanks.

PR #1217, opened 2024-09-29
improved kafka actions reading

### LINKED ISSUE #1215: Improve reading of Kafka events with receive action
**Citrus Version**
4.3.2

### High Level
In the citrus-kafka module of citrusframework, it's possible to consume (recive) events from a kafka topic. The mechanism of this recive-action however is as simple as: Consume exactly one event at the current offset of the consumer-group.

When testing an application that publishes events, it's likely that lots of events get published in a short period of time. That makes it almost impossible to find the wanted event with this mechanism.

As seen in the example in the following diagram, reading latest nor earliest will potentially suffice.

kafka-citrus drawio

### Technical
To test if an Event has been published to a Kafka Topic, isn't as simple as it might seem.

It's best to choose a consumer-group different from the SUT for testing, to not interfere with the offset of the SUT's consumer-group. But usually, when a Topic is read by a new consumer-group, it starts from the earliest or latest event published. Given that a topic might already have lots of already published events, starting with earliest would be absurdly inefficient. Initializing the consumer-group with latest on the other hand will normally result in no event read at all. Because of the chronological order of a test - i.E. an action is executed, after that we'd expect an event published to Kafka.

## Solution
The recive action could start consume a topic on an offset _O<sub>x</sub> = O<sub>T-n</sub>_. Where _n_ is the max timespan in which the wanted event is expected to have been published, but also as small as possible. A new consumer-group should then be initialized at offset _O<sub>x</sub>_ and consume until latest or until the wanted event has been found. Since the value n might be different from use-case to use-case, it should be configurable (per testcase or at least per project).

 

## Acceptance Criteria
A Kafka event, that isn't the latest nor earliest on a given topic, can be verified in a citrus test using a single citrus-action
The offset range is configurable (preferably by time)
 

## Further info
A Proof-of-concept is already done, where the `KafkaConsumer.recieve` Method looks something like this:

```java
@Override
public Message receive(TestContext context, long timeout) {
    String topic = context.replaceDynamicContentInString(Optional.ofNullable(endpointConfiguration.getTopic())
        .orElseThrow(() -> new CitrusRuntimeException("Missing Kafka topic to receive messages from - add topic to endpoint configuration")));

    var partitions = consumer.partitionsFor(topic).stream()
        .map(partition -> new TopicPartition(topic, partition.partition()))
        .toList();
    consumer.assign(partitions);

    offsetConsumerOnTopicBySeconds(partitions, 5 * 60);

    var messages = new ArrayList<Message>();
    for (int i = 0; i < 10; i++) {
        ConsumerRecords<Object, Object> records = consumer.poll(Duration.ofMillis(timeout));
        if (records.isEmpty()) {
            break;
        }
        Message received = endpointConfiguration.getMessageConverter().convertInbound(
            records.iterator().next(),
            endpointConfiguration,
            context
        );
        messages.add(received);
        context.onInboundMessage(received);
    }

    consumer.commitSync(Duration.ofMillis(endpointConfiguration.getTimeout()));

    if (messages.isEmpty()) {
        throw new MessageTimeoutException(timeout, topic);
    }
    return messages.get(0);
}

private void offsetConsumerOnTopicBySeconds(List<TopicPartition> partitions, int secondsToSubtract) {
    Map<TopicPartition, Long> partitionWithTimestamp = partitions.stream().collect(toMap(
        Function.identity(),
        partition ->  now().minusSeconds(secondsToSubtract).toEpochMilli()
    ));

    var newOffsets = consumer.offsetsForTimes(partitionWithTimestamp);
    newOffsets.forEach((partition, partitionOffset) -> {
        if (partitionOffset != null) {
            consumer.seek(partition, partitionOffset.offset() - 1);
        } else {
            consumer.seekToEnd(List.of(partition));
        }
    });
}
```

bbortt:
improved the `receive` action for Kafka with selectors. that makes it possible to look out for specific events. sample in java:

```java
public class KafkaConsumerIT extends TestNGCitrusSpringSupport {

    @Autowired
    private KafkaEndpoint helloKafkaEndpoint;

    @Test
    @CitrusTest
    public void findKafkaEvent_headerEquals_citrus_DSL() {
        var body = "findKafkaEvent_headerEquals_citrus_DSL";

        var key = "Name";
        var value = "Bilbo";

        when(
                send(helloKafkaEndpoint)
                        .message(new KafkaMessage(body).setHeader(key, value))
        );

        then(
                receive(helloKafkaEndpoint)
                        .selector(
                                kafkaConsumerSelector()
                                        .eventLookbackWindow(Duration.ofSeconds(1L))
                                        .kafkaConsumerRecordMatcher(kafkaHeaderEquals(key, value))
                                        .build()
                        )
                        .getMessageBuilderSupport()
                        .body(body)
        );
    }
}
```

xml sample:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<spring:beans xmlns="http://www.citrusframework.org/schema/testcase"
              xmlns:spring="http://www.springframework.org/schema/beans"
              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
              xsi:schemaLocation="http://www.springframework.org/schema/beans 
                                  http://www.springframework.org/schema/beans/spring-beans.xsd 
                                  http://www.citrusframework.org/schema/testcase 
                                  http://www.citrusframework.org/schema/testcase/citrus-testcase.xsd">
    <variables>
      <variable name="correlationId" value="citrus:randomNumber(10)"/>
      <variable name="messageId" value="citrus:randomNumber(10)"/>
      <variable name="traceId" value="citrus:randomString(32)"/>
      <variable name="spanId" value="citrus:randomString(16)"/>
      <variable name="user" value="bbortt"/>
    </variables>

    <actions>
      <receive endpoint="helloKafkaEndpoint">
        <description>Receive Kafka request: Kafka broker -> Citrus</description>
        <selector>
          <element name="filter-identifier" value="KafkaConsumerRecordHeaderMatcher"/>
          <element name="header-filter-key" value="traceparent"/>
          <element name="header-filter-value" value="${traceId}"/>
          <element name="header-filter-comparator" value="CONTAINS"/>
          <element name="event-lookback-window" value="PT1S"/>
        </selector>
        <message>
          <data>
            <![CDATA[
              <HelloRequest xmlns="http://citrusframework.org/schemas/samples/HelloService.xsd">
                 <MessageId>${messageId}</MessageId>
                 <CorrelationId>${correlationId}</CorrelationId>
                 <User>${user}</User>
                 <Text>Hello Citrus</Text>
               </HelloRequest>
            ]]>
          </data>
        </message>
        <header>
          <element name="Operation" value="sayHello"/>
          <element name="citrus_kafka_topic" value="hello"/>
          <element name="citrus_kafka_partition" value="0"/>
          <element name="citrus_kafka_offset" value="@assertThat(greaterThanOrEqualTo(0))@"/>
        </header>
      </receive>
    </actions>
  </testcase>
</spring:beans>
```

coverage on code:

2024-09-30-101131_3840x2160_scrot

closes #1215.

**edit:** updated the coverage report. I forgot that TestNG requires tests to be `public` 😉 
**edit 2:** added xml sample.

### DIFF
--- a/core/citrus-base/src/test/java/org/citrusframework/util/StringUtilsTest.java
+++ b/core/citrus-base/src/test/java/org/citrusframework/util/StringUtilsTest.java
@@ -1,0 +1,56 @@
+package org.citrusframework.util;
+
+import static org.citrusframework.util.StringUtils.hasText;
+import static org.citrusframework.util.StringUtils.isEmpty;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class StringUtilsTest {
+
+    @DataProvider
+    public static Object[][] text() {
+        return new Object[][]{{"foo"}};
+    }
+
+    @DataProvider
+    public static Object[][] blankText() {
+        return new Object[][]{{" "}};
+    }
+
+    @DataProvider
+    public static Object[][] emptyText() {
+        return new Object[][]{{null}, {""}};
+    }
+
+    @Test(dataProvider = "text")
+    public void hasText_returnsTrue(String str) {
+        assertTrue(hasText(str));
+    }
+
+    @Test(dataProvider = "blankText")
+    public void hasText_returnsFalse_forBlankText(String str) {
+        assertFalse(hasText(str));
+    }
+    @Test(dataProvider = "emptyText")
+    public void hasText_returnsFalse_forEmptyText(String str) {
+        assertFalse(hasText(str));
+    }
+
+    @Test(dataProvider = "emptyText")
+    public void isEmpty_returnsTrue(String str) {
+        assertTrue(isEmpty(str));
+    }
+
+    @Test(dataProvider = "text")
+    public void isEmpty_returnsFalse_forText(String str) {
+        assertFalse(isEmpty(str));
+    }
+
+    @Test(dataProvider = "blankText")
+    public void isEmpty_returnsFalse_forBlankText(String str) {
+        assertFalse(isEmpty(str));
+    }
+}
--- a/endpoints/citrus-kafka/pom.xml
+++ b/endpoints/citrus-kafka/pom.xml
@@ -79,21 +79,8 @@
     </dependency>
     
     <dependency>
-      <groupId>io.netty</groupId>
-      <artifactId>netty-handler</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.netty</groupId>
-      <artifactId>netty-transport-native-epoll</artifactId>
-      <classifier>linux-x86_64</classifier>
-    </dependency>
-    <dependency>
       <groupId>org.apache.zookeeper</groupId>
       <artifactId>zookeeper</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.dropwizard.metrics</groupId>
-      <artifactId>metrics-core</artifactId>
     </dependency>
 
     <dependency>
@@ -127,5 +114,4 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
 </project>
--- a/endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageConsumerUtils.java
+++ b/endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageConsumerUtils.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.kafka.endpoint;
+
+import lombok.NoArgsConstructor;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.citrusframework.context.TestContext;
+import org.citrusframework.exceptions.CitrusRuntimeException;
+import org.citrusframework.message.Message;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Optional;
+
+import static lombok.AccessLevel.PRIVATE;
+
+@NoArgsConstructor(access = PRIVATE)
+final class KafkaMessageConsumerUtils {
+
+    private static final Logger logger = LoggerFactory.getLogger(KafkaMessageConsumerUtils.class);
+
+    static String resolveTopic(
+            KafkaEndpointConfiguration kafkaEndpointConfiguration,
+            TestContext testContext
+    ) {
+        return testContext.replaceDynamicContentInString(
+                Optional.ofNullable(kafkaEndpointConfiguration.getTopic())
+                        .orElseThrow(() -> new CitrusRuntimeException(
+                                "Missing Kafka topic to receive messages from - add topic to endpoint configuration")));
+    }
+
+    static Message parseConsumerRecordsToMessage(
+            List<ConsumerRecord<Object, Object>> consumerRecords,
+            KafkaEndpointConfiguration endpointConfiguration,
+            TestContext testContext
+    ) {
+         if (consumerRecords.size() > 1) {
+            throw new CitrusRuntimeException("More than one matching record found in topic " + resolveTopic(endpointConfiguration, testContext));
+        }
+
+        if (logger.isDebugEnabled()) {
+            consumerRecords.forEach(
+                    record -> logger.debug("Received message: ({}, {}) at offset {}", record.key(), record.value(), record.offset()));
+        }
+
+        Message received = endpointConfiguration.getMessageConverter()
+                .convertInbound(
+                        consumerRecords.iterator().next(),
+                        endpointConfiguration,
+                        testContext);
+        testContext.onInboundMessage(received);
+
+        return received;
+    }
+}
--- a/endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageFilteringConsumer.java
+++ b/endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageFilteringConsumer.java
@@ -1,0 +1,226 @@
+/*
+ * Copyright the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.kafka.endpoint;
+
+import jakarta.annotation.Nullable;
+import lombok.Builder;
+import lombok.Getter;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.TopicPartition;
+import org.citrusframework.context.TestContext;
+import org.citrusframework.exceptions.CitrusRuntimeException;
+import org.citrusframework.exceptions.MessageTimeoutException;
+import org.citrusframework.kafka.endpoint.selector.KafkaMessageSelector;
+import org.citrusframework.message.Message;
+import org.citrusframework.messaging.AbstractSelectiveMessageConsumer;
+import org.slf4j.Logger;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+
+import static java.lang.String.format;
+import static java.time.Instant.now;
+import static java.util.Collections.singletonList;
+import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
+import static java.util.concurrent.Executors.newSingleThreadExecutor;
+import static java.util.stream.Collectors.toMap;
+import static org.apache.commons.lang3.exception.ExceptionUtils.getRootCause;
+import static org.citrusframework.kafka.endpoint.KafkaMessageConsumerUtils.parseConsumerRecordsToMessage;
+import static org.citrusframework.kafka.endpoint.KafkaMessageConsumerUtils.resolveTopic;
+import static org.citrusframework.util.StringUtils.hasText;
+import static org.citrusframework.util.StringUtils.isEmpty;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * A specialized Kafka message consumer that filters messages based on specified criteria.
+ * <p>
+ * This consumer provides functionality to receive Kafka messages that match given selectors or matchers within a
+ * specified time window.
+ *
+ * @see KafkaMessageSelector
+ */
+@Getter
+class KafkaMessageFilteringConsumer extends AbstractSelectiveMessageConsumer {
+
+    private static final Logger logger = getLogger(KafkaMessageFilteringConsumer.class);
+
+    private final ExecutorService executor = newSingleThreadExecutor();
+
+    private final org.apache.kafka.clients.consumer.KafkaConsumer<Object, Object> consumer;
+
+    private KafkaMessageFilter kafkaMessageFilter;
+
+    @Builder
+    private KafkaMessageFilteringConsumer(
+            KafkaEndpointConfiguration endpointConfiguration,
+            org.apache.kafka.clients.consumer.KafkaConsumer<Object, Object> consumer,
+            @Nullable Duration eventLookbackWindow,
+            @Nullable Duration pollTimeout,
+            @Nullable KafkaMessageSelector kafkaMessageSelector
+    ) {
+        super(KafkaMessageSingleConsumer.class.getSimpleName(), endpointConfiguration);
+
+        this.consumer = consumer;
+
+        if (nonNull(eventLookbackWindow) || nonNull(kafkaMessageSelector) || nonNull(pollTimeout)) {
+            kafkaMessageFilter = KafkaMessageFilter.kafkaMessageFilter()
+                    .eventLookbackWindow(eventLookbackWindow)
+                    .kafkaMessageSelector(kafkaMessageSelector)
+                    .pollTimeout(pollTimeout)
+                    .buildFilter();
+        }
+    }
+
+    @Override
+    public Message receive(String selector, TestContext testContext, long timeout) {
+        if (isEmpty(selector) && (isNull(kafkaMessageFilter))) {
+            throw new CitrusRuntimeException("Cannot invoke filtering kafka message consumer without selectors");
+        } else if (hasText(selector)) {
+            kafkaMessageFilter = KafkaMessageFilter.kafkaMessageFilter(selector);
+        }
+
+        kafkaMessageFilter.sanitize();
+
+        String topic = resolveTopic(getEndpointConfiguration(), testContext);
+
+        logger.debug("Receiving Kafka message on topic '{}' using selector: {}", topic, kafkaMessageFilter);
+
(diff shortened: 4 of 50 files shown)


endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/matcher/KafkaConsumerRecordMatcher.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.citrusframework.kafka.endpoint.matcher;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.citrusframework.message.MessageSelectorBuilder;
+
+import java.util.Map;
+
+/**
+ * Interface for defining custom matching logic for finding {@link ConsumerRecord}'s in Kafka. Implementations of this
```
tschlat, line 25:
```suggestion
 * Defines custom matching logic for filtering {@link ConsumerRecord} instances in Kafka. Implementations of this
```

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/endpoint/KafkaConsumerSelectorTest.java
```diff
@@ -0,0 +1,240 @@
...
+    @Test
+    public void hasKafkaConsumerRecordMatcherFactory() {
+        assertThat(fixture.getKafkaConsumerRecordMatcherFactory())
+                .isNotNull();
+    }
+
+    @Test
+    public void defaultPollTimeout() {
+        assertThat(fixture.getPollTimeout())
+                .isEqualTo(Duration.ofMillis(100));
```
tschlat, line 57: Seems to me like a very short time for default poll period. But i might be wrong.

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/integration/KafkaEndpointJavaIT.java
```diff
@@ -0,0 +1,232 @@
...
+                        .message(new KafkaMessage(body).setHeader(key, value))
+        );
+
+        then(
+                kafkaWithRandomConsumerGroupEndpoint.findKafkaEventHeaderEquals(Duration.ofSeconds(1L), key, value)
+                        .body(body)
+        );
+    }
+
+    @Configuration
```
tschlat, line 221:
```suggestion
    @TestConfiguration
```

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/endpoint/KafkaConsumerTest.java
```diff
@@ -157,6 +188,6 @@ public void testWithMessageHeaders() {
         Message receivedMessage = endpoint.createConsumer().receive(context);
         Assert.assertEquals(receivedMessage.getPayload(), controlMessage.getPayload());
         Assert.assertNotNull(receivedMessage.getHeader("Operation"));
-        Assert.assertTrue(receivedMessage.getHeader("Operation").equals("sayHello"));
+        Assert.assertEquals(receivedMessage.getHeader("Operation"), "sayHello");
```
tschlat, line 191: static imports?

tschlat (changes requested):

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/integration/KafkaEndpointJavaIT.java
bbortt: that's strangely not available here. I can double-check, but it might need an additional module.

bbortt: yeah well, that's part of `spring-boot`. not sure if the additional dependency is worth it. wdym?

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/endpoint/KafkaConsumerTest.java
bbortt: didn't want to refactor too much. but done.

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/endpoint/KafkaConsumerSelectorTest.java
bbortt: it's the time it takes to read one message, not the general poll timeout. I've even thought about just setting this statically. no clue if it's reasonable, but from what I've read it seems to be.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageFilteringConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+                    .eventLookbackWindow(eventLookbackWindow)
+                    .kafkaConsumerRecordMatcher(kafkaConsumerRecordMatcher)
+                    .pollTimeout(pollTimeout)
+                    .buildSelector();
+        }
+    }
+
+    @Override
+    public Message receive(String selector, TestContext testContext, long timeout) {
+        if (isEmpty(selector) && (isNull(kafkaConsumerSelector))) {
```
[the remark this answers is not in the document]
bbortt: it's either the `selector` or the `kafkaConsumerSelector`. `selector` is for citrus DSL, `kafkaConsumerSelector` for java DSL,

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageConsumerUtils.java
```diff
@@ -0,0 +1,69 @@
...
+            List<ConsumerRecord<Object, Object>> consumerRecords,
+            ParserConfiguration parserConfiguration
+    ) {
+         if (consumerRecords.size() > 1) {
+            throw new CitrusRuntimeException("More than one matching record found in topic " + resolveTopic(parserConfiguration.endpointConfiguration, parserConfiguration.testContext));
+        }
+
+        if (parserConfiguration.logger.isDebugEnabled()) {
+            consumerRecords.forEach(
+                    record -> parserConfiguration.logger.debug("Received message: ({}, {}) at offset {}", record.key(), record.value(), record.offset()));
```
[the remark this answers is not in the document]
bbortt: it's a utils class.. don't know. I could.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageSingleConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+            consumer.subscribe(Arrays.stream(topic.split(",")).toList());
+        }
+
+        ConsumerRecords<Object, Object> consumerRecords = consumer.poll(Duration.ofMillis(timeout));
+        if (consumerRecords.isEmpty()) {
+            throw new MessageTimeoutException(timeout, topic);
+        }
+
+        var received = parseConsumerRecordsToMessage(
+                StreamSupport.stream(consumerRecords.spliterator(), false).toList(),
```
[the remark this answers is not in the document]
bbortt: no, there are two `stream`-s in use here. one of `Arrays#stream` and `StreamSupport#stream`.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageSingleConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+    public Message receive(TestContext testContext, long timeout) {
+        String topic = resolveTopic(getEndpointConfiguration(), testContext);
+
+        logger.debug("Receiving Kafka message on topic: '{}'", topic);
+
+        if (consumer.subscription() == null || consumer.subscription().isEmpty()) {
+            consumer.subscribe(Arrays.stream(topic.split(",")).toList());
+        }
+
+        ConsumerRecords<Object, Object> consumerRecords = consumer.poll(Duration.ofMillis(timeout));
```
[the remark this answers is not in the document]
bbortt: yes, but I preferred to see what's actually invoked.. `ofMillis` could be anything. ok if I leave it as it is?

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaProducer.java
```diff
@@ -16,81 +16,91 @@
...
 
         try {
             ProducerRecord<Object, Object> producerRecord = endpointConfiguration.getMessageConverter().convertOutbound(message, endpointConfiguration, context);
-            producer.send(producerRecord).get(endpointConfiguration.getTimeout(), TimeUnit.MILLISECONDS);
-            logger.info("Message was sent to Kafka stream topic: '" + topic + "'");
+            producer.send(producerRecord).get(endpointConfiguration.getTimeout(), MILLISECONDS);
+            logger.info("Message was sent to Kafka stream topic: '{}'", topic);
         } catch (InterruptedException | ExecutionException e) {
-            throw new CitrusRuntimeException(String.format("Failed to send message to Kafka topic '%s'", topic), e);
+            throw new CitrusRuntimeException(format("Failed to send message to Kafka topic '%s'", topic), e);
```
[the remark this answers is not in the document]
bbortt: I think keeping those together is ok. both are of `java.util.concurrent`. I've flagged the thread however.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageFilteringConsumer.java
```diff
@@ -0,0 +1,224 @@
...
+        final Future<List<ConsumerRecord<Object, Object>>> handler = executor.submit(() -> findMessagesSatisfyingMatcher(topic));
+
+        try {
+            return handler.get(timeout, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException | InterruptedException | ExecutionException e) {
+            logger.error("Failed receiving Kafka message: {}", getRootCause(e).getMessage());
+
+            handler.cancel(true);
+
+            if (Thread.interrupted()) {
```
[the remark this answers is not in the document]
bbortt: there is no difference between `Thread.interrupted()` or `Thread.currentThread().isInterrupted()`.

core/citrus-base/src/test/java/org/citrusframework/util/StringUtilsTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.testng.Assert.assertTrue;
+
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class StringUtilsTest {
+
+    @DataProvider
+    public static Object[][] text() {
+        return new Object[][]{{"foo"}};
```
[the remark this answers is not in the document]
bbortt: yes, but the name is very readable.. I've removed it.

endpoints/citrus-kafka/pom.xml
```diff
@@ -78,30 +78,30 @@
...
-      <groupId>io.dropwizard.metrics</groupId>
-      <artifactId>metrics-core</artifactId>
-    </dependency>
 
     <dependency>
       <groupId>jakarta.xml.bind</groupId>
       <artifactId>jakarta.xml.bind-api</artifactId>
     </dependency>
 
+    <dependency>
```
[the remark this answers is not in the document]
bbortt: well, it would harm the compile-time I guess. but I don't have a strong opinion on that.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaConsumerSelector.java
```diff
@@ -0,0 +1,160 @@
...
+import static java.util.Objects.nonNull;
+import static lombok.AccessLevel.PACKAGE;
+import static lombok.AccessLevel.PRIVATE;
+
+@Getter(PACKAGE)
+@AllArgsConstructor(access = PRIVATE)
+@RequiredArgsConstructor(access = PRIVATE)
+@Builder(buildMethodName = "buildSelector")
+@ToString(exclude = {"kafkaConsumerRecordMatcherFactory"})
+public final class KafkaConsumerSelector {
```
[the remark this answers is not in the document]
bbortt: I've thought really long and deep about it (previous to your comment), but I didn't like to expose the whole `Map` interface. the keys that are useable are fixed. there seems no point to me.

with the current implementation (and the builder) the java API is really nicely done, in my opinion.

I think the name makes sense, however.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageConsumerUtils.java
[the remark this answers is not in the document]
tschlat: It's where the error occurs - right?

endpoints/citrus-kafka/src/test/java/org/citrusframework/kafka/integration/KafkaEndpointJavaIT.java
bbortt: update: there is even a strange context-loading problem when taking spring boot onto the classpath. I won't do it.

endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageFilteringConsumer.java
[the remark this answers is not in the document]
tschlat: There is a difference
  
```java
public static boolean interrupted() {
     return currentThread().getAndClearInterrupt();
}
    
public boolean isInterrupted() {
    return interrupted;
}
    ```
As you see, the static method clears the interrupted state, which is not what you want. Instead you should actually catch the interrupted exception and make sure, the current thread gets interrupted. This is common practice and ensures that the state of the thread can be acknowledged further up the call hierarchy.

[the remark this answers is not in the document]
Comment (bbortt) [reply endpoints/citrus-kafka/src/main/java/org/citrusframework/kafka/endpoint/KafkaMessageFilteringConsumer.java:143]: ups, sorry. didn't read it careful enough. used the wrong method 😢

File: src/manual/endpoint-kafka.adoc [hunk 14]
```diff
@@ -103,20 +103,22 @@ public KafkaEndpoint helloKafkaEndpoint() {
...
-topic. As the communication is asynchronous by default producer does not wait for a synchronous response.
+The endpoint is now ready to be used inside a test case.
+The test simply references the endpoint by its name when sending or receiving.
 
-In case of a receive operation the endpoint creates a Kafka consumer instance in the defined consumer group. The consumer
-starts a subscription on the given topic and acts as a polling listener. This means that the message consumer connects
-to the given topic and polls for records. As soon as a record has been received the action is ready to perform the validation
-process that verifies the record.
+In case of a send operation the endpoint creates a Kafka producer and will simply publish the records to the defined Kafka topic.
+As the communication is asynchronous by default producer does not wait for a synchronous response.
```
valisol, line 113:
```suggestion
As the communication is asynchronous by the default producer, the producer does not wait for a synchronous response.
```

src/manual/endpoint-kafka.adoc
```diff
@@ -103,20 +103,22 @@ public KafkaEndpoint helloKafkaEndpoint() {
...
 
-NOTE: By default, the consumer polls for a single record per receive operation (`max.poll.records=1`). You can change this
-setting within the consumer properties on the Citrus Kafka endpoint. The property `max.poll.records` that sets the number
-of records in a single poll.
+In case of a receive operation the endpoint creates a Kafka consumer instance in the defined (possibly random) consumer group.
+
+NOTE: By default, the consumer starts a subscription on the given topic and acts as a polling listener.
+This means that the message consumer connects to the given topic and polls one single record at a time.
+As soon as a record has been received the action is ready to perform the validation process that verifies the record.
```
valisol, line 119:
```suggestion
As soon as a record has been received, the action is ready to perform the validation process that verifies the record.
```

src/manual/endpoint-kafka.adoc
```diff
@@ -103,20 +103,22 @@ public KafkaEndpoint helloKafkaEndpoint() {
...
+This requires looping for finding specific Kafka messages and is potentially very ineffective, given the Kafka topic receives a high amount of traffic.
```
valisol, line 120: Würde den Satz ein bisschen anders gestalten.
```suggestion
This approach requires looping to find specific Kafka messages, which can be highly inefficient, especially when the Kafka topic experiences heavy traffic.
```

src/manual/endpoint-kafka.adoc
```diff
@@ -189,52 +190,63 @@ The following table shows all available settings on a Kafka endpoint in Citrus:
...
 | No
-| citrus_kafka_group
-| Consumer group name. Please keep in mind that records are load balanced across consumer instances with the same consumer
-  group name set. So you might run into message timeouts when using multiple Kafka endpoints with the same consumer group name.
+| `citrus_kafka_group`
+| Consumer group name.
+Please keep in mind that records are load balanced across consumer instances with the same consumer group name set.
+So you might run into message timeouts when using multiple Kafka endpoints with the same consumer group name.
+
+| random-consumer-group
```
valisol, line 198:
```suggestion
| `random-consumer-group`
```

bbortt: I've corrected this for all properties, not only the random consumer group.

PR #1221, opened 2024-10-01
fix(citrus-kafka): delombok module

bbortt:
as a cleanup for https://github.com/citrusframework/citrus/pull/1217.

cc: @tschlat we had a little discussion going on about lombok and decided not to use it for the moment. mainly because some companies have stricter rules than we do regarding lombok and prohibit the use of it or any libraries that include it. we will ask the community for their input at a given time.


christophd (approved):
many thanks

bbortt:
by the way, coverage is still premium!

image

PR #1223, opened 2024-10-03
chore: remove author annotations from java sources

bbortt:
bulk change to remove all authors. applied regex: `.*@author.*\n`, restricted to `*.java* files only.

then looked out for now-empty comments:

```java
/**
*/
```

and removed with another bulk change.


christophd (approved):
LGTM Sorry I have caused merge conflicts

PR #1237, opened 2024-10-17
fix: disallow access into ws interceptors

bbortt:
**problem**; access to `interceptors` list is exposed to public. modifications to it won't be propagated to the underlying web service template.

I personally find this a bit irritating. I think the acessor methods could be used instead, that's what this PR would change.

see [this (unchanged) setter implementation](https://github.com/citrusframework/citrus/pull/1237/files#diff-e9300a011f3d44e61f8d81ccc215c373a1ba8d4163639f765e914e79d5dbd06eR240):

```java
    public void setInterceptors(List<ClientInterceptor> interceptors) {
        this.interceptors = interceptors;
        getWebServiceTemplate().setInterceptors(interceptors.toArray(new ClientInterceptor[0]));
    }
```

### DIFF
--- a/endpoints/citrus-ws/src/main/java/org/citrusframework/ws/client/WebServiceEndpointConfiguration.java
+++ b/endpoints/citrus-ws/src/main/java/org/citrusframework/ws/client/WebServiceEndpointConfiguration.java
@@ -16,12 +16,13 @@
 
 package org.citrusframework.ws.client;
 
+import static java.util.Collections.singletonList;
+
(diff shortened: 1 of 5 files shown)


christophd:
could you please elaborate why this has been an issue, thx

bbortt:
> could you please elaborate why this has been an issue, thx

@christophd updated the PR description.

christophd (approved):
many thanks

Issue #929, opened 2023-03-03
LoggingClientInterceptor do not print German Umlauts


chokdee:
**Citrus Version**
3.4.0

**Expected behavior**
Printing German Umlauts with correct encoding

**Actual behavior**
Printing wrong characters to log

**Test case sample**
handleRequest(getRequestContent(request, new String(body)));

should be 
handleRequest(getRequestContent(request, new String(body, StandardCharsets.UTF_8)));
or even better, use the encoding from the http header

bbortt:
I agree, it should: Respect encoding from HTTP headers or fallback to UTF-8.

PR #1247, opened 2024-10-30
Issue #929: fix set charset for request content

phos-web:
Get encoding charset from HTTP headers or fallback to UTF-8

### DIFF
--- a/endpoints/citrus-http/src/main/java/org/citrusframework/http/interceptor/LoggingClientInterceptor.java
+++ b/endpoints/citrus-http/src/main/java/org/citrusframework/http/interceptor/LoggingClientInterceptor.java
@@ -32,12 +32,13 @@ import org.springframework.http.client.ClientHttpResponse;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.StandardCharsets;
+import java.io.UnsupportedEncodingException;
 import java.util.List;
 import java.util.Map.Entry;
 
 import static java.lang.String.join;
 import static java.lang.System.lineSeparator;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Simple logging interceptor writes Http request and response messages to the console.
@@ -120,6 +121,24 @@ public class LoggingClientInterceptor implements ClientHttpRequestInterceptor {
      * @return
      */
     private String getRequestContent(HttpRequest request, String body) {
+        String contentType = request.getHeaders().getFirst(HttpHeaders.CONTENT_TYPE);
+        if (contentType != null) {
+            String[] contentTypeParts = contentType.split(";");
+            for (String contentTypePart : contentTypeParts) {
+                if(contentTypePart.startsWith("charset=") && !contentTypePart.endsWith("charset=")) {
+                    String charset = contentTypePart.split("=")[1];
+                    try {
+                        body = new String(body.getBytes(), charset);
+                    } catch (UnsupportedEncodingException e) {
+                        body = new String(body.getBytes(), UTF_8);
+                    }
+                }
+                break;
+            }
+        } else {
+            body = new String(body.getBytes(), UTF_8);
+        }
+
         StringBuilder builder = new StringBuilder();
 
         builder.append(request.getMethod());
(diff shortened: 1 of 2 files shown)


endpoints/citrus-http/src/test/java/org/citrusframework/http/interceptor/LoggingClientInterceptorTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    public void testGetRequestContentWithNullContentType() throws Exception {
+        HttpRequest request = Mockito.mock(HttpRequest.class);
+        HttpHeaders headers = Mockito.mock(HttpHeaders.class);
+
+        when(request.getMethod()).thenReturn(HttpMethod.valueOf("GET"));
+        when(request.getURI()).thenReturn(new URI("http://example.com"));
+        when(request.getHeaders()).thenReturn(headers);
+        when(headers.getFirst(HttpHeaders.CONTENT_TYPE)).thenReturn(null);
```
bbortt, line 76: try this as another test case, please: `when(headers.getFirst(HttpHeaders.CONTENT_TYPE)).thenReturn("");`

endpoints/citrus-http/src/main/java/org/citrusframework/http/interceptor/LoggingClientInterceptor.java
```diff
@@ -120,6 +121,24 @@ public boolean hasMessageListeners() {
...
+        String contentType = request.getHeaders().getFirst(HttpHeaders.CONTENT_TYPE);
+        if (contentType != null) {
+            String[] contentTypeParts = contentType.split(";");
+            for (String contentTypePart : contentTypeParts) {
+                if(contentTypePart.startsWith("charset=")) {
+                    String charset = contentTypePart.split("=")[1];
+                    try {
+                        body = new String(body.getBytes(), charset);
+                    } catch (UnsupportedEncodingException e) {
+                        body = new String(body.getBytes(), StandardCharsets.UTF_8);
```
bbortt, line 133: please use static imports wherever possible.

```suggestion
                        body = new String(body.getBytes(), UTF_8);
```

endpoints/citrus-http/src/main/java/org/citrusframework/http/interceptor/LoggingClientInterceptor.java
```diff
@@ -120,6 +121,24 @@ public boolean hasMessageListeners() {
      * @return
      */
     private String getRequestContent(HttpRequest request, String body) {
+        String contentType = request.getHeaders().getFirst(HttpHeaders.CONTENT_TYPE);
+        if (contentType != null) {
+            String[] contentTypeParts = contentType.split(";");
+            for (String contentTypePart : contentTypeParts) {
+                if(contentTypePart.startsWith("charset=")) {
+                    String charset = contentTypePart.split("=")[1];
```
bbortt, line 129: if for some reason the header contains only `charset=`? this line will fail. please add a check `contentTypePart.split("=").length == 2`.

bbortt (changes requested):
thank you. I think one or two more checks would be valid. but it shouldn't be far from the final solution.

closes https://github.com/citrusframework/citrus/issues/929.

endpoints/citrus-http/src/test/java/org/citrusframework/http/interceptor/LoggingClientInterceptorTest.java
phos-web: added this as another test case called `testGetRequestContentWithEmptyStringContentType()`

endpoints/citrus-http/src/main/java/org/citrusframework/http/interceptor/LoggingClientInterceptor.java
phos-web: added static imports

endpoints/citrus-http/src/main/java/org/citrusframework/http/interceptor/LoggingClientInterceptor.java
phos-web: prevent failure by adding `!contentTypePart.endsWith("charset=")` to if clause.

bbortt (approved):

PR #1251, opened 2024-11-03
chore: Add Testcontainers connector module

christophd:
- Provide test actions to interact with Testcontainers
- Add AWS2 LocalStack container support
- Add PostgreSQL container support
- Add XML DSL support
- Add YAML DSL support


connectors/citrus-testcontainers/src/main/java/org/citrusframework/testcontainers/TestContainersSettings.java
```diff
@@ -0,0 +1,121 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.testcontainers;
+
+import java.util.Optional;
+
+public class TestContainersSettings {
```
bbortt, line 21: if this is just a utility class, it could be `final`:

```suggestion
public final class TestContainersSettings {
```

I think utility classes are a good use-case for this.

connectors/citrus-testcontainers/src/main/java/org/citrusframework/testcontainers/TestcontainersHelper.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.testcontainers;
+
+public class TestcontainersHelper {
```
bbortt, line 19: same goes here regarding `final` classes.

bbortt (commented):
that's a nice addition. just a thought regarding utility-/settings-classes (there are quit a few of those in citrus). up to you.

PR #1254, opened 2024-11-06
feat: add json array order configuration option in java

novarx:


validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonElementValidatorTest.java
```diff
@@ -258,6 +250,64 @@ public static JsonAssertion[] validOnlyWithIgnoreExpressions() {
         ).toArray(new JsonAssertion[0]);
     }
 
+    @Test
+    void validate_shouldCheckArrayOrder_whenPropertyTrue() {
```
bbortt, line 254: these must be `public` or they won't be executed.

bbortt (changes requested):
just one note on the tests.

bbortt:
I think this is obsolete @novarx @tschlat

Issue #1258, opened 2024-11-07
ci: Build hanging due to endless loop in Kafka tests


christophd:
**Citrus Version**
latest

**Description**
The GitHub actions workflow is running into endless loop while verifying as part of the Kafka tests. I think this is due to running tests in parallel and recently I have introduced Testcontainers module that also start some Kafka tests. Need to investigate.

Asa first step we should find a proper loop ending condition on the tests to avoid the build running forever.

Examples:
https://github.com/citrusframework/citrus/actions/runs/11710427055
https://github.com/citrusframework/citrus/actions/runs/11710751209

bbortt:
@christophd I've started working on this a week ago already! my idea was to introduce a timeout to the "try until condition" (respectively the base class of these loop-action-types). I think that would be a nice feature anyway. because, as you can see, if the condition isn't met; it just loops forever.

I have the implementation ready, just need to transfer it into the base class. didn't find time since last week, but would sure be able to complete it on saturday.

christophd:
that sounds awesome! many thanks and take your time

christophd:
@bbortt I think we can close this one fixed, right?

bbortt:
I think so. didn't see it anymore.

PR #1261, opened 2024-11-09
add timeout to iterating action containers

bbortt:
as promised, a timeout implementation on all iterating actions. this is currently:
* repeat until condition
* repeat on error
* iterate

this will partially help with https://github.com/citrusframework/citrus/issues/1258, because the kafka IT's won't run forever.


bbortt:
wll, that works: https://github.com/citrusframework/citrus/actions/runs/11757100929/job/32753963702?pr=1261#step:6:16585.

PR #1282, opened 2024-12-17
kafka concurrent access

### LINKED ISSUE #1281: KafkaMessageFilteringConsumer#findMessageWithTimeout concurrent access
**Citrus Version**
4.4

**Expected behavior**
KafkaMessageFilteringConsumer should poll messages without running into exceptions.

**Actual behavior**
Consumer runs into the following exception:
```
org.citrusframework.exceptions.TestCaseFailedException: KafkaConsumer is not safe for multi-threaded access. currentThread(name: main, id: 1) otherThread(id: 50)
```

The Reason is, that the conusmer is queried asynchronously here:
KafkaMessageFilteringConsumer#findMessageWithTimeout:

```
Future<List<ConsumerRecord<Object, Object>>> handler = executorService.submit(() -> {
    return this.findMessagesSatisfyingMatcher(topic);
});
```

and then here from the main thread:
```
        } finally {
            executorService.shutdownNow();
            this.consumer.unsubscribe();
        }

```

When the consumber is unsubscribed, it is still locked by the asynchonous thread. So before unsubscription, we need a sync here to make sure the consumer is not in use any more.

**Test case sample**
> Please, share the test case (as small as possible) which shows the issue

bbortt:
closes #1281.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/messaging/AbstractMessageConsumer.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/messaging/AbstractMessageConsumer.java
@@ -34,7 +34,7 @@ public abstract class AbstractMessageConsumer implements Consumer {
     /**
      * Default constructor using receive timeout setting.
      */
-    public AbstractMessageConsumer(String name, EndpointConfiguration endpointConfiguration) {
+    protected AbstractMessageConsumer(String name, EndpointConfiguration endpointConfiguration) {
         this.name = name;
         this.endpointConfiguration = endpointConfiguration;
     }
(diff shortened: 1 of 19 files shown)


bbortt:
> @[bbortt](https://github.com/bbortt) bbortt requested review from [tschlat](https://github.com/tschlat) and [christophd](https://github.com/christophd) and removed request for [tschlat](https://github.com/tschlat)

ah yea, that makes sense 💃🏼 my clicking is way too fast for GitHub.

tschlat (changes requested):

bbortt:
@tschlat I've now added a flag `useThreadSafeConsumer`... that makes it backwards compatible.

christophd (approved):

PR #1283, opened 2024-12-18
fix(deps): updated dependency versions

bbortt:
noteable breaking change included:

* `io.fabric8` from 6.13.4 to 7.0.1


christophd (approved):
LGTM many thanks

Issue #1290, opened 2025-01-08
Kubernetes actor unit test is flaky in GH actions CI


christophd:
The Kubernetes actor unit test interferes with other tests running in parallel in GH actions workflow. This is because the test forcibly sets the disabled state on the Kubernetes clients to verify the disabled state.

Other tests running in parallel may be influenced by this setting, too.

Make sure to run this test in single threaded mode.

PR #1291, opened 2025-01-08
fix(1290): Fix flaky Kubernetes actor tests

christophd:
- Make sure to run the Kubernetes actor unit test in single threaded mode in order to not influence other tests running in parallel
- Make sure to properly reset the actor's connected state after the tests


christophd:
Fixes #1290

Issue #1292, opened 2025-01-08
Revert to single threaded mode in GH actions CI


christophd:
The GH actions CI jobs fail very often due to running unit tests in parallel. Consider reverting back to single threaded execution and stabilized the concurrent test execution in a separate PR

PR #1293, opened 2025-01-08
fix(#1290): Fix default Kubernetes actor

christophd:
- Use default Kubernetes test actor only when feature is enabled


christophd:
Related to #1292

Issue #1297, opened 2025-01-16
Add Escape Json Function


flumiii:
**User story**
>As a tester, I want to use JSON return values in string variables so that they are fully preserved and not partially lost because Java applies String interpretation rules and ends the string before the end because of unescaped double quotes.

**Additional context**
> the function should be tested and documented

PR #1298, opened 2025-01-16
feat(function): add EscapeJsonFunction, test and doc

flumiii:
Add escapeJson() function and test. Update Citrus doc.

### DIFF
--- a/src/manual/functions.adoc
+++ b/src/manual/functions.adoc
@@ -549,15 +549,21 @@ If you want to deal with escaped XML in your test case you may want to use this 
 If you want to deal with escaped JSON in your test case you may want to use this function. It automatically escapes all JSON special characters (especially double quotes).
 [source,xml]
 ----
-<echo>
-    <message>
-        <![CDATA[
-            citrus:escapeJson("{\"exampleJson\": \"exampleValue\"}")
-        ]]>
-    </message>
-</echo>
+ <kafkaproxy:send-create-event topic="${kafkaTopicName}">
+                <kafkaproxy:body>
+                    <kafkaproxy:data>
+                        {
+                            "kafkaClientConfiguration": {
+                                "bootstrapServers": ["${kafkaBootstrapServer}"]
+                            },
+                            "messageBody": "citrus:escapeJson("{"exampleJson": "exampleValue"}")"
+                        }
+                    </kafkaproxy:data>
+                </kafkaproxy:body>
+                <kafkaproxy:receive responseCode="201" />
+            </kafkaproxy:send-create-event>
+
 ----
-`{"exampleJson": "exampleValue"}`
 
 [[functions-cdata-section]]
 == cdataSection()
--- a/src/manual/functions.adoc
+++ b/src/manual/functions.adoc
@@ -544,6 +544,21 @@ If you want to deal with escaped XML in your test case you may want to use this 
 
 *&lt;Message&gt;Hallo Test Framework&lt;/Message&gt;*
 
+[[functions-escape-json]]
+== escapeJson()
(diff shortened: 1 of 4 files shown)


flumiii:
https://github.com/citrusframework/citrus/issues/1297
#1297

src/manual/functions.adoc
```diff
@@ -544,6 +544,21 @@ If you want to deal with escaped XML in your test case you may want to use this
...
 
+[[functions-escape-json]]
+== escapeJson()
+If you want to deal with escaped JSON in your test case you may want to use this function. It automatically escapes all JSON special characters (especially double quotes).
+[source,xml]
+----
+<echo>
+    <message>
+        <![CDATA[
+            citrus:escapeJson("{\"exampleJson\": \"exampleValue\"}")
```
tschlat, line 555: It would be usefull to have a reasonable example here which shows the use case.
Something like

```suggestion
            { 
              "messageId": "1234",
              "messageEncodedAsJsonObject": {"exampleJson": "exampleValue"}
              "messageEncodedAsJsonString": "citrus:escapeJson('{\"exampleJson\": \"exampleValue\"}')"
            }
```

validation/citrus-validation-json/pom.xml
```diff
@@ -72,5 +72,10 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
```
tschlat, line 75: Up to my knowledge, Citrus core currently uses TestNG .

validation/citrus-validation-json/src/main/java/org/citrusframework/functions/core/EscapeJsonFunction.java
```diff
@@ -0,0 +1,31 @@
...
+ * <p>If the input is invalid (null, empty, or contains more than one string), an {@link IllegalArgumentException} will be thrown.</p>
+ *
+ * <p>Example input: <code>"{\"mySuperJson\": \"valium\"}"</code></p>
+ * <p>Example output: <code>"{\\\"mySuperJson\\\": \\\"valium\\\"}"</code></p>
+ *
+ */
+public class EscapeJsonFunction implements Function {
+    @Override
+    public String execute(List<String> list, TestContext testContext) {
+        if (Objects.isNull(list) || list.size() != 1 || list.get(0).isEmpty()) {
```
tschlat, line 24: Still a list with  a null elementcould be passed in here. So better check this:
```suggestion
        if (Objects.isNull(list) || list.size() != 1 || StringUtils.isEmpty(list.get(0))) {
```

tschlat (changes requested):

src/manual/functions.adoc
flumiii: I added an expample from the kafka-proxy test, which should work.

validation/citrus-validation-json/pom.xml
flumiii: removed dependency and rewrote parameterized tests

validation/citrus-validation-json/src/main/java/org/citrusframework/functions/core/EscapeJsonFunction.java
flumiii: added the null-ckeck which is null safe;)

src/manual/functions.adoc
flumiii: tested if example works - everything fine.

bbortt:
thanks @flumiii, I've added copyright headers. all good now.

bbortt (approved):

bbortt:
closes https://github.com/citrusframework/citrus/issues/1297

PR #1317, opened 2025-02-14
fix: Fix duplicate invocation of deprecated onTestFinish callback

christophd:
Does it make sense to invoke the deprecated `onTestFinish()` in new `onTestExecutionStart()`? I guess onTestFinish method would be called twice as the new `onTestExecutionEnd()` does also call the deprecated method, too.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/report/TestListener.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/report/TestListener.java
@@ -36,13 +36,11 @@ public interface TestListener {
     }
 
     /**
-     * Invoked when test execution has ended (after final actions execution and
-     * before {@link org.citrusframework.container.AfterTest} execution)
-     *
-     * @see #onTestEnd(TestCase)
+     * Invoked when test execution starts
+     * (after {@link org.citrusframework.container.BeforeTest} execution} execution)
      */
     default void onTestExecutionStart(TestCase test) {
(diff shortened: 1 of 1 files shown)


core/citrus-api/src/main/java/org/citrusframework/report/TestListener.java
```diff
@@ -42,7 +42,7 @@ default void onTestFinish(TestCase test) {
      * @see #onTestEnd(TestCase)
      */
     default void onTestExecutionStart(TestCase test) {
-        onTestFinish(test);
+        // Default implementation does nothing
```
tschlat, line 45: 😵

This was a last minute change using copy paste. The correct documentation and implementation should be:
(sorry for not being able to suggest on multiple lines - do not apply as patch)
```suggestion
    /**
     * Invoked when test execution starts (after {@link org.citrusframework.container.BeforeTest} execution)
     */
    default void onTestExecutionStart(TestCase test) {
        // Default implementation does nothing
    }
```

tschlat (changes requested):

core/citrus-api/src/main/java/org/citrusframework/report/TestListener.java
christophd: thank you, I have fixed the javadocs, too. as you have suggested

PR #1318, opened 2025-02-17
Issue/1175/core citrus changes ts

tschlat:
This MR contains changes to base citrus modules in preparation of OpenAPI module enhancements.

The MR contains 15 meaningful commits, so you might want to review them commit by commit.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
@@ -54,9 +54,7 @@ import org.citrusframework.validation.context.HeaderValidationContext;
 import org.citrusframework.validation.context.ValidationContext;
 import org.citrusframework.validation.json.JsonMessageValidationContext;
 import org.citrusframework.validation.json.JsonPathMessageValidationContext;
-import org.citrusframework.validation.script.ScriptValidationContext;
 import org.citrusframework.validation.xml.XmlMessageValidationContext;
-import org.citrusframework.validation.xml.XpathMessageValidationContext;
 import org.citrusframework.variable.VariableExtractor;
 import org.citrusframework.variable.dictionary.DataDictionary;
 import org.slf4j.Logger;
@@ -245,14 +243,12 @@ public class ReceiveMessageAction extends AbstractTestAction {
                 }
             } else {
                 boolean mustFindValidator = validationContexts.stream()
-                        .anyMatch(item -> JsonPathMessageValidationContext.class.isAssignableFrom(item.getClass()) ||
-                                XpathMessageValidationContext.class.isAssignableFrom(item.getClass()) ||
-                                ScriptValidationContext.class.isAssignableFrom(item.getClass()));
+                        .anyMatch(ValidationContext::requiresValidator);
 
-                List<MessageValidator<? extends ValidationContext>> validators =
+                List<MessageValidator<? extends ValidationContext>> activeValidators =
                         context.getMessageValidatorRegistry().findMessageValidators(messageType, message, mustFindValidator);
 
-                for (MessageValidator<? extends ValidationContext> messageValidator : validators) {
+                for (MessageValidator<? extends ValidationContext> messageValidator : activeValidators) {
                     messageValidator.validateMessage(message, controlMessage, context, validationContexts);
                 }
             }
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
@@ -418,7 +418,7 @@ public class ReceiveMessageAction extends AbstractTestAction {
     /**
      * Action builder.
      */
-    public static final class Builder extends ReceiveMessageActionBuilder<ReceiveMessageAction, ReceiveMessageActionBuilderSupport, Builder> {
+    public static class Builder extends ReceiveMessageActionBuilder<ReceiveMessageAction, ReceiveMessageActionBuilderSupport, Builder> {
 
         /**
          * Fluent API action building entry method used in Java DSL.
--- a/core/citrus-base/src/main/java/org/citrusframework/functions/core/AdvancedRandomNumberFunction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/functions/core/AdvancedRandomNumberFunction.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.functions.core;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.ThreadLocalRandom;
+
+import org.citrusframework.context.TestContext;
+import org.citrusframework.exceptions.CitrusRuntimeException;
+import org.citrusframework.exceptions.InvalidFunctionUsageException;
+import org.citrusframework.functions.Function;
+
+import static java.lang.String.format;
+
+/**
+ * A function for generating random double values with specified decimal places and range. This
+ * function includes options to specify the number of decimal places, minimum and maximum values,
+ * and whether to include or exclude the minimum and maximum values.
+ * <p>
+ * Parameters:
+ * <ol>
+ *     <li>Decimal places: The number of decimal places in the generated random number (optional, default: 0). Note that definition of 0 results in an integer.</li>
+ *     <li>Min value: The minimum value for the generated random number (optional, default: {@link #DEFAULT_MIN_VALUE}).</li>
+ *     <li>Max value: The maximum value for the generated random number (optional, default: {@link #DEFAULT_MAX_VALUE}).</li>
+ *     <li>Exclude min: Whether to exclude the minimum value (optional, default: false).</li>
+ *     <li>Exclude max: Whether to exclude the maximum value (optional, default: false).</li>
+ *     <li>Multiple of: The generated number will be a multiple of this value (optional).</li>
+ *     <li>Format pattern: The format pattern for the returned string (optional, default: null)</li
+ * </ol>
+ * <p>
+ */
+public class AdvancedRandomNumberFunction implements Function {
+
+    private static final BigDecimal DEFAULT_MAX_VALUE = new BigDecimal(1000000);
+    private static final BigDecimal DEFAULT_MIN_VALUE = DEFAULT_MAX_VALUE.negate();
+
+    public String execute(List<String> parameterList, TestContext context) {
+        if (parameterList == null) {
+            throw new InvalidFunctionUsageException("Function parameters must not be null.");
+        }
+
+        int decimalPlaces = getParameter(parameterList, 0, Integer.class, Integer::parseInt, 2);
+        if (decimalPlaces < 0) {
+            throw new InvalidFunctionUsageException(
+                "Decimal places must be a non-negative integer value.");
+        }
+
+        BigDecimal minValue = getParameter(parameterList, 1, BigDecimal.class, BigDecimal::new, DEFAULT_MIN_VALUE);
+        BigDecimal maxValue = getParameter(parameterList, 2, BigDecimal.class, BigDecimal::new, DEFAULT_MAX_VALUE);
+        if (minValue.compareTo(maxValue) > 0) {
+            throw new InvalidFunctionUsageException("Min value must be less than max value.");
+        }
+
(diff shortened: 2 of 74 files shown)


endpoints/citrus-http/src/main/java/org/citrusframework/http/controller/HttpMessageController.java
```diff
@@ -130,6 +134,7 @@ private ResponseEntity<?> handleRequestInternal(HttpMethod method, HttpEntity<?>
 
         HttpServletRequest servletRequest = ((ServletRequestAttributes) attributes).getRequest();
         UrlPathHelper pathHelper = new UrlPathHelper();
+        pathHelper.setRemoveSemicolonContent(!endpointConfiguration.isHandleSemicolonPathContent());
```
christophd, line 137: Instead of negating the setting maybe we should also use `isRemoveSemicolonPathContent()` on the endpoint and set this to `true` by default.

endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpEndpointConfiguration.java
```diff
@@ -102,6 +103,10 @@ public class HttpEndpointConfiguration extends AbstractPollableEndpointConfigura
      */
     private boolean disableRedirectHandling = false;
 
+    /** Should handle http semicolon uri content e.g. for matrix encoded path variables  */
+    // TODO: Christoph Deppisch advise whether this is a reasonable approach to support matrix encoded path parameters
```
christophd, line 107: seems reasonable to me. you can remove the TODO

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -418,7 +418,7 @@ public MessageBuilder getMessageBuilder() {
     /**
      * Action builder.
      */
-    public static final class Builder extends ReceiveMessageActionBuilder<ReceiveMessageAction, ReceiveMessageActionBuilderSupport, Builder> {
+    public static class Builder extends ReceiveMessageActionBuilder<ReceiveMessageAction, ReceiveMessageActionBuilderSupport, Builder> {
```
christophd, line 421: I don't see why the removal of `final` is required by this commit but I guess it is a preparation for upcoming changes. Please double check if this is required in upcoming PRs

src/manual/functions.adoc
```diff
@@ -255,10 +255,92 @@ Function output:
...
+- Possible values: UPPERCASE, LOWERCASE, MIXED.
+
+3. Digits characters (0-9) included. (Optional, default: false)
+
+- If set to true, digits will be part of the output.
+
+4. Minimum number of characters in the generated string. (Optional, default: length)
+
+- This value must be positive or zero and less than or equal to the total length.
+- If smaller than the length, the generated string will have a random length between the minimum and the specified length.
```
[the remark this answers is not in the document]
tschlat: Please note that I introduced this fourth parameter as a new feature. The function still produces the same output as before if the additional parameter is not used, ensuring backward compatibility.

This parameter allows for randomness in string length, which is useful for random OpenAPI string attribute generation, particularly when a minimum length constraint applies. In many cases, a variable-length string within an allowed range can be more suitable than a fixed-length string.

Since the function remains non-breaking and retains its original behavior when the fourth parameter is omitted, I don’t see how this introduces an issue.

If strict fixed-length behavior is required, I could introduce a separate function (e.g., randomStringWithRandomLength), but I find this approach unnecessary and redundant.

tschlat (commented):
There is just one open discussion left.

src/manual/functions.adoc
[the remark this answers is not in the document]
christophd: yes, thanks for the discussion. I am aware that there is no issue with backward compatibility. I just find it weird that an additional parameter implicitly changes the behavior of the function. previously this was fixed length only.

But overall I am not on a strong opinion here. the documentation on the parameters is good so I a fine keeping it in `randomString()` function

tschlat:
@bbortt do you want to have a closer look or can we continue?

src/manual/functions.adoc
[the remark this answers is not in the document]
tschlat: Then, if you are fine with it, i would prefer keeping it as it is.

bbortt:
> @bbortt do you want to have a closer look or can we continue?

I will, hang on...

core/citrus-base/src/main/java/org/citrusframework/functions/core/AdvancedRandomNumberFunction.java
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * A function for generating random double values with specified decimal places and range. This
+ * function includes options to specify the number of decimal places, minimum and maximum values,
+ * and whether to include or exclude the minimum and maximum values.
+ * <p>
+ * Parameters:
+ * <ol>
+ *     <li>Decimal places: The number of decimal places in the generated random number (optional, default: 0). Note that definition of 0 results in an integer.</li>
+ *     <li>Min value: The minimum value for the generated random number (optional, default: Double.MIN_VALUE).</li>
+ *     <li>Max value: The maximum value for the generated random number (optional, default: Double.MAX_VALUE).</li>
```
bbortt, line 40: these are not the actual `DEFAULT_MAX_VALUE` and `DEFAULT_MIN_VALUE` being used (see https://github.com/citrusframework/citrus/pull/1318/files#diff-167bf384ca89ef57c71db3fb84f936a1a26b31d840611638192d2ce6e97f525cR54), or am I being misleaded? if so, the variable naming could be improved, I think. I am also wondering why these `DEFAULT_*` variables must be `public`?

core/citrus-base/src/main/java/org/citrusframework/functions/core/RandomStringFunction.java
```diff
@@ -18,21 +18,23 @@
...
+import static java.lang.System.currentTimeMillis;
 
 /**
  * Function generating a random string containing alphabetic characters. Arguments specify
  * upper and lower case mode.
  *
  */
 public class RandomStringFunction implements Function {
-    private static Random generator = new Random(System.currentTimeMillis());
+    private static final Random generator = new Random(currentTimeMillis());
```
bbortt, line 37:
```suggestion
    private static final Random RANDOM = new Random(currentTimeMillis());
```

convention for `static final` variables.

core/citrus-base/src/test/java/org/citrusframework/functions/core/AdvancedRandomNumberFunctionTest.java
```diff
@@ -0,0 +1,411 @@
...
+        assertEquals(exception.getMessage(),
+            "Function parameters must not be null.");
+    }
+
+    @Test
+    public void testRandomNumberWithDefaultValues() {
+        List<String> params = List.of();
+        String result = function.execute(params, context);
+        assertNotNull(result);
+        assertTrue(result.matches("-?\\d*\\.\\d{2}"));
```
bbortt, line 57: just a hint, but `assertJ` could help with things like `isGreaterThanOrEqualTo` and `isLessThanOrEqualTo` to check the bounds. because, as I highlighted in another comment, I think the function is not correctly bound to `Double.MAX_VALUE` and `Double.MIN_VALUE`.

I would also love this to be repeated tests, because the function returns something with a _probability_. you can invoke this using `@Test(invocationCount = 1000)`, I think that should be sufficient. same goes for all tests in this class.

core/citrus-base/src/test/java/org/citrusframework/util/StringUtilsTest.java
```diff
@@ -54,4 +58,152 @@ public void isEmpty_returnsFalse_forText(String str) {
     public void isEmpty_returnsFalse_forBlankText(String str) {
         assertFalse(isEmpty(str));
     }
+
+    @Test
+    public void appendSegmentToPath() {
+        assertEquals(StringUtils.appendSegmentToUrlPath("s1", "s2"), "s1/s2");
```
bbortt, line 64:
```suggestion
        assertEquals(appendSegmentToUrlPath("s1", "s2"), "s1/s2");
```

`quote` is also being imported statically, but this not. is this intentional?

core/citrus-spring/src/main/java/org/citrusframework/config/xml/SendMessageActionParser.java
```diff
@@ -131,7 +104,17 @@ protected BeanDefinitionBuilder parseComponent(Element element, ParserContext pa
      */
     public static class SendMessageActionFactoryBean extends AbstractSendMessageActionFactoryBean<SendMessageAction, SendMessageAction.SendMessageActionBuilderSupport, SendMessageAction.Builder> {
 
-        private final SendMessageAction.Builder builder = new SendMessageAction.Builder();
+        private final SendMessageAction.Builder builder;
+
```
bbortt, line 108:
```suggestion
```

endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpQueryParamHeaderValidator.java
```diff
@@ -38,49 +42,62 @@ public void validateHeader(String name, Object received, Object control, TestCon
...
-            super.validateHeader(HttpMessageHeaders.HTTP_QUERY_PARAMS, received, control, context, validationContext);
+            super.validateHeader(HTTP_QUERY_PARAMS, received, control, context, validationContext);
             return;
         }
 
-        Map<String, String> receiveParams = convertToMap(received);
-        Map<String, String> controlParams = convertToMap(control);
+        // TODO Christoph Deppisch: I changed this to support multi lists, e.q. required for array query parameters.
+        //  Not sure about consequences though. Therefore, i call a new method super.validateHeaderArray below.
+        //  Maybe we should fix this in general.
```
bbortt, line 51: shoutout to @christophd, did you see (and approve of) this?

src/manual/functions.adoc
```diff
@@ -269,6 +352,7 @@ This function will generate a random string representation with a defined length
     <variable name="rndString3" value="citrus:randomString(10, LOWERCASE)"/>
     <variable name="rndString4" value="citrus:randomString(10, MIXED)"/>
     <variable name="rndString4" value="citrus:randomString(10, MIXED, true)"/>
+    <variable name="rndString4" value="citrus:randomString(10, MIXED, true,2)"/>
```
bbortt, line 355:
```suggestion
    <variable name="rndString4" value="citrus:randomString(10, MIXED, true, 2)"/>
```

src/manual/functions.adoc
```diff
@@ -281,8 +365,41 @@ AgSSwedetG
...
+ABCDE
+735-RT
+bcda42
+TgH56yJk12
+----
+
+The generated string strictly adheres to the provided regex pattern.
+The function supports character ranges, quantifiers, and grouping as per standard Java regex syntax.
+If an invalid regex is provided, an error will be thrown. An error might also be thrown for very complex
+regex, which cannot be handled by the Generex library.
```
bbortt, line 401:
```suggestion
If an invalid regex is provided, an error will be thrown.
An error might also be thrown for very complex regex, which cannot be handled by the Generex library.
```

endpoints/citrus-http/src/main/java/org/citrusframework/http/message/HttpQueryParamHeaderValidator.java
christophd: Honestly, I am not sure about the consequences neither. All existing tests are green so that drives my confidence

bbortt: love that mindset/confidence 🥇

core/citrus-base/src/main/java/org/citrusframework/functions/core/RandomStringFunction.java
tschlat: It's not really a constant for which i chose lower case. . You can always set a new seed. I'd keep it in lower case but I have no really strong opinion on that. I applied the name suggestion though.

core/citrus-base/src/main/java/org/citrusframework/functions/core/AdvancedRandomNumberFunction.java
tschlat: I set these default values to avoid generation of incredibly large numbers by default. Still it is possible to set min/max to higher values. Internally BigDecimal is used as value. The naming in the javadoc is wrong. I adjusted it. Also made the constants private.

core/citrus-base/src/test/java/org/citrusframework/functions/core/AdvancedRandomNumberFunctionTest.java
tschlat: It is not bound to these values. I just set some defaults as explained above.

I set an invocation count where appropriate. Internally there have been loops to accomodate for this "random issue". Might have missed some though. Anyway, this class now produces 8800 Tests.

I adjusted some tests and code. So you might want to take a further look.

core/citrus-spring/src/main/java/org/citrusframework/config/xml/SendMessageActionParser.java
tschlat: :shipit:

core/citrus-base/src/main/java/org/citrusframework/functions/core/RandomStringFunction.java
bbortt: I see the thought of that. it's constant in terms of "one runtime", right? neither have I a strong opinion, in that case.

core/citrus-base/src/test/java/org/citrusframework/functions/core/AdvancedRandomNumberFunctionTest.java
bbortt: still passes in around 103 ms. all good imo.

bbortt (approved):

PR #1337, opened 2025-04-09
fix(citrus-base): assume most common message types before validation

bbortt:
**edit:** moved bug documentation to: https://github.com/citrusframework/citrus/issues/1338.

closes https://github.com/citrusframework/citrus/issues/1338.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
@@ -16,16 +16,6 @@
 
 package org.citrusframework.actions;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -217,18 +224,20 @@ protected void validateMessage(Message message, TestContext context) {
...
         }
 
-        if (validationProcessor != null) {
+        if (nonNull(validationProcessor)) {
             validationProcessor.validate(message, context);
         } else {
             logger.debug("Control message:\n{}", controlMessage.print(context));
 
-            if (!validators.isEmpty()) {
+            assumeMessageType(message);
```
bbortt, line 238: actually, I think it should be on `controlMessage` if it exists.

```suggestion
            if (nonNull(controlMessage)) {
              assumeMessageType(controlMessage);
            } else {
              assumeMessageType(message);
            }
```

bbortt:
@christophd now this one is good to go. I initially forgot that `XHTML` is a `MessageType` as well (and is also XML-Like).

christophd (approved):
LGTM thanks!

PR #1348, opened 2025-05-25
Issue/1347: Set `System.out` to a log4j-managed `PrintStream`

### LINKED ISSUE #1347: citrus-agent: logs to stdout are not captured by log4j
Currently, data written to `System.out` is not captured by log4j and is therefore 
 
- not formatted properly, and 
- not present in the log file. 

This is especially noticeable when using the testNG engine since [testNG prints its final report to `System.out`](https://github.com/testng-team/testng/blob/d50b2ad2d6809d52131a07071fe229b1b901e08c/testng-core/src/main/java/org/testng/SuiteRunnerWorker.java#L68-L89).

It would be nice if we were able to "redirect" everything printed to `System.out` to log4j.

**User story**
>As a user of citrus,
> When something is written to `System.out`
> Then I want this to be redirected to log4j
> So that it can be formatted properly and appears in the log file.

turing85:
Resolves #1347

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1068,6 +1068,11 @@
       </dependency>
       <dependency>
         <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-iostreams</artifactId>
+        <version>${log4j2.version}</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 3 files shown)


christophd (approved):
Looks good, many thanks!

Issue #1345, opened 2025-05-15
OSSRH service end of life


christophd:
Maven Central OSSRH service is going end of life in June 30th 2025.

This means we need to adjust the Maven release process and migrate to publishing with the new Central Portal https://central.sonatype.org/publish/publish-portal-guide/

PR #1349, opened 2025-05-26
fix(#1345): Update to Maven central portal release

christophd:
- Use Maven central portal instead of deprecated OSSRH plugins
- Adjust release scripts to publish to Maven Central portal


christophd:
Fixes #1345

PR #1366, opened 2025-07-03
fix: receive message type assumption on nullish values

bbortt:
a too eager message type assumption using `getPayload(String)` on the message leads to `null`'is values not being thus. e.g. it's interpreted as `XML` instead of `PLAINTEXT`. occured during implementation of https://github.com/citrusframework/citrus-simulator/pull/315.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
@@ -16,16 +16,6 @@
 
 package org.citrusframework.actions;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 import org.citrusframework.CitrusSettings;
 import org.citrusframework.context.TestContext;
 import org.citrusframework.endpoint.Endpoint;
@@ -68,8 +58,19 @@ import org.citrusframework.variable.dictionary.DataDictionary;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
+import static java.util.stream.Collectors.joining;
 import static org.citrusframework.message.MessageType.mapToMessageType;
 import static org.citrusframework.util.StringUtils.hasText;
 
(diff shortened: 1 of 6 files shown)


core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -84,52 +84,80 @@
...
 
     /**
      * The expected message type to arrive in this receive action.
      * This information is needed to find a proper message validator for this message
-     * */
+     */
     private String messageType;
 
-    /** Logger */
+    /**
```
tschlat, line 158: This doc could as well be removed ;-)

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -289,31 +320,30 @@ protected void validateMessage(Message message, TestContext context) {
...
         }
 
         if (MessageType.isBinary(getMessageType()) ||
-            MessageType.FORM_URL_ENCODED.equalsIgnoreCase(getMessageType())) {
+                MessageType.FORM_URL_ENCODED.equalsIgnoreCase(getMessageType())) {
             return;
         }
 
-        var payload = message.getPayload(String.class);
+        var payload = isNull(message.getPayload()) && "null".equals(message.getPayload(String.class)) ? null : message.getPayload(String.class);
```
tschlat, line 346: Can't this be done better? It is smelly. Having a "null" string here seems to be the issue. Can't this be solved earlier? In my opinion message.getPayload(String.class) should never return "null" unless there is a payload which really contains "null"

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -336,11 +366,14 @@ private void assumeMessageType(Message message) {
 
                 setMessageType(MessageType.YAML);
             }
+        } else {
+            setMessageType(MessageType.PLAINTEXT);
```
tschlat, line 370: Since this is not really abvious, it might be worth a comment. I think it is done as such because there is always a plain message validator by default?

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -336,11 +366,14 @@ private void assumeMessageType(Message message) {
...
         }
     }
 
     /**
      * Create control message that is expected. Apply global and local message processors and data dictionaries.
+     *
```
tschlat, line 376: All this javadoc is useless. We used to remove such zero information doc in the past. You may want to stick with that.

tschlat (approved):

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt: good one.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt: done. class cleaned up.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt: I think the problem was something along the line of this: a JSON message with `null`'ish payload resolves to `{}` when calling `Message#getPayload(String)` - but of course it does not have an `Object` representation, because it is an empty object. it is not "effectively nullish". thus a check on `Message#getPayload()` is not sufficient. it is additionally required to make sure the String representation is also "null".

I've refactored the code a bit, moved this into a method with a comment. that should make it more clear.

I don't think it can be catched earlier on, respectively this would go to deep and change too much of "accepted behavior".

bbortt (commented):
I've amended your remarks @tschlat! thanks.

christophd:
would it help to set the default message type to PLAINTEXT? I think XML as a default is a historic value and as you have mentioned there is always a default validator for PLAINTEXT available.

If it helps to simplify the checks we can change the default to PLAINTEXT. Otherwise I am also good with the solution provided as it is right now. Many thanks!

bbortt:
phu, we've tried (@tschlat and me) but it goes quit deep.. I fell like that change would be too risky.

thanks for the review!

PR #1388, opened 2025-08-13
Allow DTD as we are in controlled environment

stefan-ziel:
Groovy by default blocks XML with DTD this brakes tests of legacy systems that still work with DTD


bbortt (commented):
@christophd I do not see any harm in this. the constructor is as following:

```java
public XmlSlurper(boolean validating, boolean namespaceAware, boolean allowDocTypeDeclaration) throws ParserConfigurationException, SAXException {
```

PR #1416, opened 2025-09-19
fix(#1415): allow "null" retrieval from

### LINKED ISSUE #1415: JsonPathSegmentVariableExtractor and XPathSegmentVariableExtractor fail to resolve "null" values
**Citrus Version**
4.8

**Expected behavior**
The extractors should successfully extract an existing null value from the source

**Actual behavior**
The extractors fail with "unresolvable" message.

**Test case sample**
> Please, share the test case (as small as possible) which shows the issue

tschlat:
- improve error message


tschlat:
> @tschlat this caused side effects which I tried to tackle. but there were too many for a Sunday evening 😉

Thanx fro trying. I fired aand forgot on friday. I will have a look at it.

Issue #1420, opened 2025-09-25
OpenAPI requests fail to add random path values and cookies


tschlat:
**Citrus Version**
4.8

**Expected behavior**
Required parameters of all kind should be filled with random values if autofill property is set to REQUIRED. 

**Actual behavior**
Cookie and path path parameters are not randomly filled.

PR #1421, opened 2025-09-25
fix(#1420): fill required path and cookie params in OpenAPI with rand…

tschlat:
…om values

- fill required path parameter with random value
- fill required cookie parameter with random value
- add respective test cases
- modify the generated test api such that all operations have a default builder method that allows for "no parameter"


bbortt (approved):
I've rebased your branch so I could test it with the latest changes and in citrusframework/citrus-simulator as well. all good! thank you very much 🚀

PR #1425, opened 2025-10-08
fix(citrus-openapi): memory leak in openapi specification repository

bbortt:
whenever you added an OpenAPI specification to the `OpenApiRepository`, the object was effectively being saved into the underlying `List`, disregarding whether this exact specifications was already registered earlier. it would leak memory after some ("enough") time,
because these objects will never be dereferences and thus, never be garbage collected.

the root cause for this is effectively the missing "contains" functionality. that functionality has now been exposed throught the `contains` method.

in addition, the `add` method is designed in an idempotent way. only "new" OpenAPI specifications will be persisted into the `List`. the comparison is based on the `getUid()` value.

and lastly,
users have the ability to reset the list using the `withOpenApiSpecifications` method.

this offers a variety of ways to resolve the problem, respecitvely fixes the memory leak.
and adds additional interaction-options for users.

### DIFF
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiRepository.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/OpenApiRepository.java
@@ -16,9 +16,6 @@
 
 package org.citrusframework.openapi;
 
-import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


bbortt:
thanks @christophd

PR #1436, opened 2025-10-31
feat(#1081): select timezone for current date

### LINKED ISSUE #1081: How to set citrus:currentDate to UTC
**Citrus Version**
Citrus 2.8.0

**Question**
Is there a way to set a currentDate to UTC?

**What I've tried so far**
In my country, I've adjusted the time by subtracting 1 hour as demonstrated below. However, within the GitLab pipeline, there's a server configured with UTC dateTime. Consequently, I'll need to modify the variable when concluding testing on my local PC…

```
<variables>
<variable name="now" value="citrus:currentDate('yyyy-MM-dd'T'HH:mm:ss'Z'','-1h')" />
</variables>
```

phos-web:
**Description**
Adds support for selecting a timezone in the currentDate function.
Users can now pass an optional timezone parameter (e.g. 'UTC', 'PST', 'Europe/Berlin') to generate dates in a specific timezone.

**Related issue**
Fixes #1081


christophd (approved):
Many thanks!

Issue #506, opened 2018-10-09
How to concat a single quote with citrus:concat?


chokdee:
**Citrus Version**
2.7.1

**Question**
How to concat a single quote with citrus:concat?

**What I've tried so far**
```
<create-variables>
   <variable  name="sql" value="citrus:concat( 'select from xx where id='', ''' )"/>
</create-variables>
```

The second value with a single quote isn't concated to the first string.
How can I do this?

**Additional information**

svettwer:
Hi!

I've investigated the issue. Currently this is a bug unfortunately. There is a pretty ugly workaround to make the single quote concatenation work use `'''''` instead of `'''`.

So you could fix your test with the provided hotfix but I'll definitely change this logic in future releases. Do you have many test cases where this issue occurs?

BR,
Sven

svettwer:
Hi!

There will be a fix for sure. I just can't provide a timberline. :wink:

BR, 
Sven

PR #1448, opened 2025-11-30
fix(#506): concat a single quote

phos-web:
**Description**
Allows concatenating with a single quote.

**Related issue**
Fixes https://github.com/citrusframework/citrus/issues/506


bbortt (changes requested):
any tests, maybe @phos-web? 😉 from what I currently can see in this PR "nothing changed". build passes, but it also did so previously. a reproducer of the issue (that fails first, then passes with the fix) would be appreciated. optionally other tests for different scenarios, imo.

PR #1457, opened 2026-01-10
Only set `this.testJar` when `testJar` parameter is not `null`

### LINKED ISSUE #1456: Setting `testJar` in `TestRunConfiguration` to `null` results in `NullPointerException`
**Citrus Version**
4.9.0

**Expected behavior**
No exception is thrown.

**Actual behavior**
Exception is thrown

**Test case sample**
Sending the following body to the `/run` endpoint of a with `citrus-remote` packaged applications results in the `NullPointerException`:

```json
{
  "engine":"testng",
  "packages":[],
  "includes":["^.*IT$","^.*ITCase$","^IT.*$"],"defaultProperties":{},
  "testJar":null,
  "verbose":true,
  "testSources":[]
}
```
Sending the following body, however, does not trigger an exception:

```json
{
  "engine":"testng",
  "packages":[],
  "includes":["^.*IT$","^.*ITCase$","^IT.*$"],"defaultProperties":{},
  "verbose":true,
  "testSources":[]
}
```

turing85:
Resolves #1456


christophd (approved):
Many thanks!

Issue #1459, opened 2026-01-14
Update Json Schema Validator Dependency to 3.0.0


valisol:
Currently Citrus is using the com.networknt.json-schema-validator dependency with version 1.5.9.

[Pom.xml](https://github.com/citrusframework/citrus/blob/main/pom.xml#L244)

Version 3.0.0 is currently out.
It would make sense to update to the newest version for validating json.

PR #1465, opened 2026-01-20
chore(#1459): load schema from SchemaLocation rather than FileInputSt…

tschlat:
Loading the schema from FileInputStream does not allow for:
- loading from Jar 
- loading schema dependencies
- loading from http

So i decided to go for loading from a SchemaLocation.


bbortt (approved):

tschlat:
> Many thanks. I have one doubt though. I think not all implementations of Resource provide a proper URL in getURL(). We should fallback to using FileInputStream when SchemaLocation fails for some reason - just to be on the safe side.

For file based resources, it looks like a file really needs to exist. Therefore taking the URL should be fine. However, to be safe i now catch any exception and  then call the previous implementation.

PR #1467, opened 2026-01-21
Draft: update to jackson 3

bbortt:


christophd:
Many thanks for your work here! It helped a lot to make #1496 happen 👍

Issue #523, opened 2018-10-26
mvn plugin - swagger test generation fails if no "200" or "default" response is given


svettwer:
**Citrus Version**
>=2.7.4

**Expected behavior**
A swagger generation containing response definitions without "200" or "default" definitions are parsed and tests are generated.

**Actual behavior**
The test execution fails completely.

**Test case sample**
The sample API https://petstore.swagger.io/v2/swagger.json contains definitions for  operations that can't be successful. E.g. the following "updatePet" Operation that just contains "400" "404" and "405" responses. No success, no default. Therefore the mojo is not able to create a test from this. 

```json
{
   "put":{
      "tags":[
         "pet"
      ],
      "summary":"Update an existing pet",
      "description":"",
      "operationId":"updatePet",
      "consumes":[
         "application/json",
         "application/xml"
      ],
      "produces":[
         "application/xml",
         "application/json"
      ],
      "parameters":[
         {
            "in":"body",
            "name":"body",
            "description":"Pet object that needs to be added to the store",
            "required":true,
            "schema":{
               "$ref":"#/definitions/Pet"
            }
         }
      ],
      "responses":{
         "400":{
            "description":"Invalid ID supplied"
         },
         "404":{
            "description":"Pet not found"
         },
         "405":{
            "description":"Validation exception"
         }
      },
      "security":[
         {
            "petstore_auth":[
               "write:pets",
               "read:pets"
            ]
         }
      ]
   }
}
```

The issue is located in `SwaggerJavaTestGenerator`, line 114 and following.

BR,
Sven

PR #1469, opened 2026-02-01
fix(#523): handle missing 200 and default response in swagger test generation

phos-web:
**Description**
Fixes the Swagger test generator crashing when an API operation defines no `200` or `default` response. The generator now falls back to the first available response in the spec and resolves its actual HTTP status code, instead of assuming a successful response always exists. If the status code cannot be resolved, the fallback remains `200 OK`.

**Related issue**
Fixes [#523](https://github.com/citrusframework/citrus/issues/523)


bbortt (changes requested):

christophd:
sorry, this PR needs a rebase due to a fix for the CI jobs provided on main branch

PR #1507, opened 2026-04-05
chore: remove dangling object mappers

bbortt:
originates from #1498.


christophd (approved):
Looks good, thanks! I think we should also adjust the documentation and search for ObjectMapper and object-mapper occurrences

bbortt:
nice catch! there were indeed mentions of those.

PR #1530, opened 2026-05-20
include most likely message validator hints

### LINKED ISSUE #1529: "Found validation context that has not been processed"
**Citrus Version**

`5.0.0-M1` (and earlier)

**Description**

The following test sends an empty request body and expects an HTTP 400 response, then validates some fields in the response body:

```java
 @Test
@CitrusTest
void shouldReturn400WhenCreatingQualityGateWithMissingRequiredFields(
  @CitrusResource TestCaseRunner testRunner
) {
  testRunner.when(
    qualityGateApi
      .sendCreateQualityGate()
      .schemaValidation(false)
      .getMessageBuilderSupport()
      .body("{}")  // empty body — all required fields absent
  );
  testRunner.then(
    qualityGateApi
      .receiveCreateQualityGate(BAD_REQUEST)
      .validate(
        jsonPath().expression("$.code", BAD_REQUEST.getReasonPhrase())
      )
  );
}
```

The test fails with "some validation context have not been processed", thrown from [ReceiveMessageAction#validateMessage](https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java#L324).

The root cause is that `citrus-validation-json` is not on the classpath — but the error message gives no indication of this, making it very difficult to diagnose.

**Suggested Fix**

Improve the exception message to include:

* The most likely validator class for the unprocessed context
* The Citrus module that provides it (e.g. `citrus-validation-json`)

This would turn a cryptic failure into an actionable one, pointing users directly to the missing dependency.

bbortt:
I suggest this solution for #1529, @christophd what do you think?

closes #1529.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/validation/context/ValidationContext.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/validation/context/ValidationContext.java
@@ -16,14 +16,14 @@
 
 package org.citrusframework.validation.context;
 
-import java.util.Map;
-import java.util.Optional;
-
 import org.citrusframework.exceptions.CitrusRuntimeException;
 import org.citrusframework.spi.ResourcePathTypeResolver;
 import org.citrusframework.spi.TypeResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+import java.util.Optional;
 
 /**
  * Basic validation context holding validation specific information.
@@ -46,6 +46,13 @@ public interface ValidationContext {
      */
     default boolean requiresValidator() {
         return false;
+    }
+
+    /**
+     * Return the most likely implementation of {@link org.citrusframework.validation.MessageValidator} that can validate this context.
+     */
+    default Optional<String> getCorrespondingValidationModule() {
+        return Optional.empty();
     }
 
     /**
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
@@ -316,13 +316,31 @@ public class ReceiveMessageAction extends AbstractTestAction implements MessageA
                 }
             }
 
-            List<ValidationContext> unknown = validationContexts.stream()
+            List<ValidationContext> validationContextsWithUnknownValidationStatus = validationContexts.stream()
                     .filter(validationContext -> validationContext.getStatus() == ValidationStatus.UNKNOWN)
                     .toList();
-            if (!unknown.isEmpty()) {
-                unknown.forEach(validationContext -> logger.warn("Found validation context that has not been processed: {}", validationContext.getClass().getName()));
-                throw new ValidationException("Incomplete message validation of message with type %s - no validation has been performed on the following contexts: %s".formatted(messageType, unknown.stream().map(validationContext -> validationContext.getClass().getSimpleName()).collect(joining(", "))));
+            if (!validationContextsWithUnknownValidationStatus.isEmpty()) {
+                validationContextsWithUnknownValidationStatus.forEach(ReceiveMessageAction::logUnusedValidationContext);
+                var unusedValidationContextString = validationContextsWithUnknownValidationStatus.stream().map(validationContext -> validationContext.getClass().getSimpleName()).collect(joining(", "));
+                throw new ValidationException(
+                        "Incomplete message validation of message with type '%s'! Conduct the log for more details.%nNo validation has been performed on the following contexts: %s".formatted(
+                                messageType,
+                                unusedValidationContextString
+                        ));
             }
+        }
+    }
+
+    private static void logUnusedValidationContext(ValidationContext validationContext) {
+        if (validationContext.getCorrespondingValidationModule().isEmpty()) {
+            logger.warn("Found validation context that has not been processed: {}", validationContext.getClass().getName());
+        } else {
+            logger.warn(
+                    "Found validation context that has not been processed: {} - you most likely forgot to include '{}' in your dependencies!",
+                    validationContext.getClass().getName(),
+                    validationContext.getCorrespondingValidationModule()
+                            .orElseThrow(IllegalStateException::new)
+            );
         }
     }
 
@@ -332,8 +350,11 @@ public class ReceiveMessageAction extends AbstractTestAction implements MessageA
             Object encoding = message.getHeaders().get("Content-Transfer-Encoding");
             MessageType messageType = mapToMessageType(
                     contentTypeString,
-                    encoding != null ? encoding.toString() : null);
-            if (messageType != null) {
+                    encoding != null ? encoding.toString() : null
+            );
+
+            if (nonNull(messageType)) {
+                logger.debug("Assumed message type '{}' for content-type: {}", messageType, contentTypeString);
                 setMessageType(messageType);
                 return;
             }
--- a/core/citrus-base/src/main/java/org/citrusframework/validation/json/JsonPathMessageValidationContext.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/validation/json/JsonPathMessageValidationContext.java
@@ -16,15 +16,16 @@
 
 package org.citrusframework.validation.json;
 
-import java.util.HashMap;
-import java.util.Map;
-
 import org.citrusframework.message.DelegatingPathExpressionProcessor;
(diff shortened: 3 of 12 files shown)


core/citrus-api/src/main/java/org/citrusframework/validation/context/ValidationContext.java
```diff
@@ -48,6 +49,13 @@ default boolean requiresValidator() {
         return false;
     }
 
+    /**
+     * Return the most likely implementation of {@link org.citrusframework.validation.MessageValidator} that can validate this context.
+     */
+    default @Nullable UnknownValidationHint getValidationHint() {
+        return null;
```
christophd: Instead of returning null by default we could use a generic hint like

```java
"Validation context `%s` that has not been processed, you most likely forgot to include a Citrus validation module.".formatted(this.getClass().getName())
```

core/citrus-api/src/main/java/org/citrusframework/validation/context/UnknownValidationHint.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.citrusframework.validation.context;
+
+public record UnknownValidationHint(
+        String mostLikelyMessageValidatorClass,
```
christophd: I do not see the `mostLikelyMessageValidatorClass` being used. I think we do always hav access to the validation context class name not being processed so we can use that instead and remove the property completely.

If so we can also think of removing the record and just using the Citrus module as a String returned by the validation context.

core/citrus-api/src/main/java/org/citrusframework/validation/context/ValidationContext.java
```diff
@@ -48,6 +49,13 @@ default boolean requiresValidator() {
         return false;
     }
 
+    /**
+     * Return the most likely implementation of {@link org.citrusframework.validation.MessageValidator} that can validate this context.
+     */
+    default @Nullable UnknownValidationHint getValidationHint() {
```
christophd: Maybe no need to provide the record UnknownValidationHint and instead just provide the Citrus module as a String? Maybe return an Optional<String> here and use Optional.empty() by default.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -316,24 +317,56 @@ protected void validateMessage(Message message, TestContext context) {
...
+    private static void logUnusedValidationContext(ValidationContext validationContext) {
+        var unknownValidationHint = validationContext.getValidationHint();
+
+        if (isNull(unknownValidationHint) || isClassLoadable(unknownValidationHint.mostLikelyMessageValidatorClass())) {
+            logger.warn("Found validation context that has not been processed: {}", validationContext.getClass().getName());
+        } else {
+            logger.warn(
+                    "Found validation context that has not been processed: {} - you most likely forgot to include '{}' in your dependencies!",
+                    validationContext.getClass().getName(),
+                    unknownValidationHint.citrusModule()
```
christophd: I think only the Citrus module information is used here occasionally. Maybe refactor this to an optional String representing the module information returned by the validation context.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -316,24 +317,56 @@ protected void validateMessage(Message message, TestContext context) {
...
+            );
+        }
+    }
+
+    private static boolean isClassLoadable(String className) {
+        try {
+            Class.forName(className, false, getSystemClassLoader());
```
christophd: There is a ClassLoaderHelper utility that should be used to get the classloader. This is used throughout the whole code base in order to allow dynamically downloaded Maven modules that should be added to the ClassLoader. Something like `ClassLoaderHelper.getClassLoader()`

core/citrus-base/src/main/java/org/citrusframework/validation/json/JsonPathMessageValidationContext.java
```diff
@@ -55,6 +56,14 @@ public boolean requiresValidator() {
         return true;
     }
 
+    @Override
+    public UnknownValidationHint getValidationHint() {
```
christophd: What about the other validation context implementations like `XMLMessageValidationContext`? Do you plan to also provide validation hints with module information to those?

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonPathMessageValidatorTest.java
```diff
@@ -314,6 +320,22 @@ public void shouldFindProperValidationContext() {
...
+    @Test
+    public void isMostLikelyMessageValidatorClass_ofJsonPathMessageValidationContext() {
+        var unknownValidationHint = new JsonPathMessageValidationContext(
+                mock(JsonPathMessageValidationContext.Builder.class)
+        ).getValidationHint();
+
+        assertThat(unknownValidationHint)
+                .isNotNull()
+                .satisfies(
+                        validationHint -> assertThat(validationHint.mostLikelyMessageValidatorClass())
```
christophd: I think this is the only place where `mostLikelyMessageValidatorClass()` is used - do we really need it? Maybe just use the validation context class name

christophd (commented):
Thanks, this makes sense as it improves the validation errors and gives detailed information how to solve the issue.

I have added some comments regarding just using the Citrus module information as we should always have access to the validation context type information

core/citrus-api/src/main/java/org/citrusframework/validation/context/UnknownValidationHint.java
bbortt: no. it's about the processor, not the context! that resides in another maven module.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt: didn't know that, thanks!

core/citrus-base/src/main/java/org/citrusframework/validation/json/JsonPathMessageValidationContext.java
bbortt: yes! this was just a draft. I wanted to get your opinion first - instead of implementing something then rewriting it all over again 🤣

core/citrus-api/src/main/java/org/citrusframework/validation/context/UnknownValidationHint.java
bbortt: but I agree that the module is probably more relevant than the exact processor. funny enough, I initially only printed the module 🤣

core/citrus-api/src/main/java/org/citrusframework/validation/context/ValidationContext.java
bbortt: I don't think it is a good idea to copy this exception message all over the code. I'd rather return a small, contained hint here. and construct the message (a standard message) in the `ReceiveMessageAction`: https://github.com/citrusframework/citrus/pull/1530/changes#diff-98b05c6a7a6ce5d2144bef6f2222e9e915591ba7af143f47dca69c66a0e0c2bdR335.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt: did do so.

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt: anyway, this is now not even necessary. I've removed the method. the "hint" is just the module name, nothing more.

validation/citrus-validation-json/src/test/java/org/citrusframework/validation/json/JsonPathMessageValidatorTest.java
bbortt: did do so.

christophd (approved):
LGTM

PR #1532, opened 2026-05-21
feat(#1531)!: change default custom validator mode from EXCLUSIVE to COMBINED

### LINKED ISSUE #1531: Change default custom validator mode from EXCLUSIVE to COMBINED
**Citrus Version**

5.0.0-M1 (targeting 5.0.0 / Matrix release)

**Context**

With the major 5.0 release underway, this is the right moment to introduce breaking changes that clean up long-standing API confusion.

The current [`EXCLUSIVE` default for custom validators](https://github.com/citrusframework/citrus/blob/cb605ff4977d8a970bcf3f45a822a455063caabf/core/citrus-api/src/main/java/org/citrusframework/CitrusSettings.java#L275) was [intentional](https://github.com/citrusframework/citrus/issues/1419#issuecomment-3332351173) — it was designed to let users substitute the default validation logic entirely with their own. However, this default causes subtle, hard-to-diagnose issues in practice.

**Problem**

When a custom validator is provided, it runs exclusively, silently bypassing all other validation. This means any other constraints passed to `validate()` are simply ignored. Two concrete (pseudo-code'ish) examples:

**Example 1** — contradictory assertions on the same field:

```java
javatestRunner.then(
  receiveMessage()
    .validate(jsonPath().expression("$.id", "foo"))
    .validate((message, testContext) -> {
      assertThat(message).contains("$.id", "bar");
    })
);
```

The JSON Path assertion (`$.id == "foo"`) is silently skipped. Only the custom validator runs.

**Example 2** — custom validator makes all other configuration irrelevant:

```java
testCaseRunner.then(
  api.receiveGetAccountDetails(NOT_FOUND)
    .validate((message, testContext) -> assertThat(message.getPayload(String.class)).isEmpty())
);
```

No matter what else is configured on `receiveGetAccountDetails(NOT_FOUND)`, the test passes as long as the custom validator does — giving a false sense of security.
Both cases led to real debugging confusion (see https://github.com/citrusframework/citrus/issues/1422).

**Suggested Fix**

Change the default custom validator mode from`EXCLUSIVE` to `COMBINED`, so that custom validators run alongside other validation rather than replacing it: https://github.com/citrusframework/citrus/blob/cb605ff4977d8a970bcf3f45a822a455063caabf/core/citrus-api/src/main/java/org/citrusframework/CitrusSettings.java#L275
Users who genuinely need the old exclusive behaviour can still opt in explicitly, but the footgun default is gone.

bbortt:
When a custom validator is provided, it previously ran exclusively, silently bypassing all other validation. This means any other constraints passed to `validate()` was simply ignored. This BREAKING CHANGE changes the default custom validator mode from `EXCLUSIVE` to `COMBINED`, so that custom validators run alongside other validation rather than replacing it.

Users who genuinely need the old exclusive behaviour can still opt in explicitly, but the footgun default is gone.

BREAKING CHANGE: Change default custom validator mode from `EXCLUSIVE` to `COMBINED`.

Closes #1531.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/CitrusSettings.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/CitrusSettings.java
@@ -272,7 +272,7 @@ public final class CitrusSettings {
 
     public static final String CUSTOM_VALIDATOR_STRATEGY_PROPERTY = "citrus.custom.validator.strategy";
     public static final String CUSTOM_VALIDATOR_STRATEGY_ENV = "CITRUS_CUSTOM_VALIDATOR_STRATEGY";
-    public static final CustomValidatorStrategy CUSTOM_VALIDATOR_STRATEGY_DEFAULT = CustomValidatorStrategy.EXCLUSIVE;
+    public static final CustomValidatorStrategy CUSTOM_VALIDATOR_STRATEGY_DEFAULT = CustomValidatorStrategy.COMBINED;
 
     /**
      * Flag to enable/disable input stream caching
--- a/core/citrus-api/src/test/java/org/citrusframework/CitrusSettingsTest.java
+++ b/core/citrus-api/src/test/java/org/citrusframework/CitrusSettingsTest.java
@@ -56,12 +56,12 @@ public class CitrusSettingsTest {
     }
 
     @Test
-    public void getCustomValidatorStrategy_shouldReturnExclusiveByDefault() {
+    public void getCustomValidatorStrategy_shouldReturnCombinedByDefault() {
         systemProperties.remove(CUSTOM_VALIDATOR_STRATEGY_PROPERTY);
         environmentVariables.remove(CUSTOM_VALIDATOR_STRATEGY_ENV);
 
         assertThat(CitrusSettings.getCustomValidatorStrategy())
-                .isEqualTo(EXCLUSIVE);
+                .isEqualTo(COMBINED);
     }
 
     @DataProvider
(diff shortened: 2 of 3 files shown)


bbortt:
actually, now I am not so sure if this is a good idea. it also forces you to add e.g. body validation in `MessageStoreTest` despite the fact, that you maybe don't want to do this:

image

christophd:
ah ok, this is not what we want. 

Is it complaining about a missing validation context validation?

bbortt:
yes. I din't have time to look further into it. but there was a hanging body validation context, I think. and some default message validators were also not being validated e.g. on HTTP DELETE requests where there is no response body at all.

maybe if we refactor the default validarion context mechanism, but anyway it isn't as straight forward as I thought.

bbortt:
I did some more thinking during dancing classes @christophd 😂 I think the most straight forward approach would be to throw an exception where there is just a warning log in place today.

having default validators _and_ custom validator in place, with EXCLUSIVE mode is simply useless. default validators won't be taken into account.

in order not to break the API we added a warning log, previously. now I think we should throw an exception, to remove the chance of any false-positives. 

or am I missing something?

christophd:
> in order not to break the API we added a warning log, previously. now I think we should throw an exception, to remove the chance of any false-positives.

So this removes the COMBINED option and enforces the EXCLUSIVE option to really be exclusive, right?

bbortt:
not "removing" (the property still exists). but forces EXCLUSIVE option to really be exclusive. I will give it a try, I think.

PR #1535, opened 2026-05-25
receive message action with wrong default validation context

bbortt:


core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
bbortt, line 792: @christophd & @tschlat I think this code block is wrong here. `getMessageType` just returns `PLAINTEXT` as default. but that does _not_ work if there is e.g. an HTTP 204 NO_CONTENT response - there is no payload.

https://github.com/citrusframework/citrus/blob/6cacee555507e3d7d4b01bf9f7d364393700c4c0/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java#L857

corresponds to https://github.com/citrusframework/citrus/issues/1536.

PR #1538, opened 2026-05-26
fix(#1536): test message validator hint for default context

### LINKED ISSUE #1536: ReceiveMessageAction fails with DefaultMessageValidationContext has not been processed for OpenAPI operations with no response body
**Citrus Version**

5.0.0-M1

**Expected behavior**

A receive action generated from an OpenAPI operation that defines no response body (e.g. 204 No Content) should pass without requiring any custom validator. There is no body to validate, so the framework should not expect any body-oriented validation context to be processed.

Test case example:

```java
  @Test
  @CitrusTest
  void shouldDeleteQualityGate(@CitrusResource TestCaseRunner testRunner) {
     testRunner.when(
      qualityGateApi.sendDeleteQualityGate(qualityGateConfig.getName())
    );
    testRunner.then(qualityGateApi.receiveDeleteQualityGate(NO_CONTENT));
  }
```

It is a test using generated citrus actions from the `citrus-openapi-codegen-maven-plugin`.
This is the OpenAPI specification for the endpoint:

```yaml
    delete:
      operationId: deleteQualityGate
      summary: Delete a Quality-Gate Configuration
      description: Deletes a Quality-Gate Configuration by name
      tags:
        - quality-gate
      responses:
        '204':
          description: Configuration deleted successfully
```

There is no response body, thus no message validation _can_ take place.

**Actual behavior**

The test fails! Instead, the following error is thrown from the `ReceiveMessageAction`:

> Found validation context that has not been processed: DefaultMessageValidationContext

The only workaround is appending a no-op custom validator:

```java
testRunner.then(qualityGateApi.receiveDeleteQualityGate(NO_CONTENT)
    .validate((m, c) -> {}));
```

**Root cause**

When `ReceiveMessageAction.validateMessage()` finishes running all validators, it checks that every `ValidationContext` in its list was marked as processed by at least one validator. For responses that have a body, this works fine — the JSON/XML validator claims the `DefaultMessageValidationContext` and marks it processed.
For body-less responses, no body-oriented validator runs, so `DefaultMessageValidationContext` is never marked processed, and the framework throws an error — even though there was nothing to validate in the first place.

The no-op `.validate((m, c) -> {})` workaround works because the custom validation processor causes Citrus to skip or satisfy the unprocessed-context check.

bbortt:
closes https://github.com/citrusframework/citrus/issues/1536. in addition to: https://github.com/citrusframework/citrus/pull/1530.

### DIFF
--- a/core/citrus-api/src/main/java/org/citrusframework/validation/context/DefaultValidationContext.java
+++ b/core/citrus-api/src/main/java/org/citrusframework/validation/context/DefaultValidationContext.java
@@ -16,17 +16,23 @@
 
 package org.citrusframework.validation.context;
 
+import java.util.Optional;
+
 /**
  * Default validation context keeps track of its status to mark if this context has been processed during the validation.
+ *
  * @since 2.4
  */
 public class DefaultValidationContext implements ValidationContext {
 
-    /** The status of this context */
+    /**
+     * The status of this context
+     */
     private ValidationStatus status = ValidationStatus.UNKNOWN;
 
     /**
      * Updates the validation status if update is allowed according to the current status.
+     *
      * @param status the new status
      */
     public void updateStatus(ValidationStatus status) {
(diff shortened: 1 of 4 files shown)


christophd (approved):
LGTM, many thanks

Issue #1447, opened 2025-11-28
Major changes in dependency landscape: spring 7, jackson 3, junit 6


bbortt:
it seems citrus is currently not spring 7 compatible, mainly regarding web dependencies. I think maintaining compatibility for multiple versions (namely 6 and 7) will be extremely hard to manage.

I've tried with the new openapi citrus action generator and run into `Nested exception is: Class org.springframework.http.HttpHeaders does not implement the requested interface org.springframework.util.MultiValueMap`. the error seems to originate from `citrus-http` dependening on `spring-web` for the `RestTemplate`.

* https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpClient.java#L129
* https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-http/pom.xml#L83

one way to resolve this would probably be to get rid of this dependency in the first place. still not an easy task.

jackson 3 is also out. this might be the easiest replacement/upgrade.

and junit 6, of course. I think junit 6 should be introduced in a new module, also not that hard.

thoughts @christophd?

bbortt:
that sounds reasonable. upgrades on the user end are hard to manage. I've tried with Spring-Boot 4 M1, M2 and finally 4.0.0 and it took me some days each time to upgrade 😓 therefore, I too think many projects and users will stay on Spring 6 for the time being.

I will start looking into it then, and create a new branch. I will try and submit the changes in reasonable pieces.

christophd:
@bbort just a heads-up. I am also going to start working on these major upgrades. I suggest to do this on main branch so we update the main branch to 5.0.0-SNAPSHOT. I would cut a new branch 4.10.x before doing that so we can release another 4.x version of Citrus if needed.

So we can target a 1st Citrus 5.0.0 release and keep iterating and improving on that one. WDYT?

bbortt:
hell yes! well done @christophd , huge task completed 😳👍🏼 do you plan to release an RC? should I test a local-built SNAPSHOT? or what's the plan?

christophd:
Thanks @bbortt yes I am planning to get a 1st milestone release out soon

vmeyer:
Thanks @christophd, i tested it in the integration within multiple Microservices, using the snapshot build in context of the springboot 4 upgrade and it seems, that all works fine so far. Do you have still have a date in mind, when the RC will be available for use?

christophd:
[...]

PR #1539, opened 2026-05-26
migrate swagger-validator-core to openapi-validator-core v3

bbortt:
migration of:

```xml
- <artifactId>swagger-request-validator-core</artifactId>
+ <artifactId>openapi-request-validator-core</artifactId>
```

see: https://github.com/atlassian/openapi-request-validator#important

`openapi-request-validator-core` [supports OpenAPI v3.1](https://github.com/atlassian/openapi-request-validator/blob/master/docs/FEATURES.md), whereas `swagger-request-validator-core` only [supported OpenAPI v3.0](https://github.com/atlassian/openapi-request-validator/blob/swagger-request-validator-2.46.1/docs/FEATURES.md).

I think, this update is quit important. [OpenAPI 3.1 has been around since 2024, with OpenAPI 3.2 also being out since 2025](https://spec.openapis.org/oas/).

### DIFF
--- a/connectors/citrus-openapi/pom.xml
+++ b/connectors/citrus-openapi/pom.xml
@@ -47,7 +47,7 @@
     </dependency>
     <dependency>
       <groupId>com.atlassian.oai</groupId>
-      <artifactId>swagger-request-validator-core</artifactId>
+      <artifactId>openapi-request-validator-core</artifactId>
     </dependency>
 
     <!--
--- a/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/random/RandomConfiguration.java
+++ b/connectors/citrus-openapi/src/main/java/org/citrusframework/openapi/random/RandomConfiguration.java
@@ -54,7 +54,7 @@ public class RandomConfiguration {
         // Note that the order of generators in the list is relevant, as the list is traversed from start to end, to find the first matching generator for a schema, and some generators match for less significant schemas.
         generators.add(new RandomEnumGenerator());
         generators.add(randomGeneratorBuilder(TYPE_STRING, FORMAT_DATE).build((randomContext, schema) -> randomContext.getRandomModelBuilder().appendSimpleQuoted("citrus:currentDate('yyyy-MM-dd')")));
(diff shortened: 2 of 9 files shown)


bbortt:
@christophd it turns out they raised the minimal build target for `openapi-request-validator-core` to JDK 21: https://github.com/atlassian/openapi-request-validator/blob/master/pom.xml#L54 - the GitHub build [confirms this](https://github.com/citrusframework/citrus/actions/runs/26454068299/job/77882820912?pr=1539). We [briefly started the discussion](https://github.com/citrusframework/citrus/issues/1447#issuecomment-4012090462) of upgrading before but let it go afterwards.

Spring still supports JDK 17-25+, as do Jackson and JUnit - otherwise citrus wouldn't compile at all. That said, looking ahead, the case for upgrading now is strong:

Supported RHEL Versions | End of Full Support | End of ELS-1
-- | -- | --
OpenJDK 11 | October 31, 2024 | October 31, 2027
OpenJDK 17 | December 31, 2027 | N/A
OpenJDK 21 | December 31, 2029 | N/A
OpenJDK 25 | December 31, 2030 | N/A

(This is [from RedHad](https://access.redhat.com/articles/1299013), but e.g. Temurin is in-sync with that.)

* The library has already forced the decision - staying on the old validator version or carrying a patch just creates more maintenance burden.
* The "later" option doesn't exist cleanly; any delay means upgrading reactively under time pressure instead of on our own terms.
* JDK 17 LTS support ends in roughly 18 months anyway, so we'd be revisiting this soon regardless.

Considering this, what do you think? I'm actually pro moving on. It reduces build overhead as well — we would only need compatibility builds for JDK 25 instead of both 21 and 25. It would also give us the opportunity to clean up the codebase a bit via migrations like pattern matching and record patterns.

**Edit:** And this...

> openapi-request-validator-core [supports OpenAPI v3.1](https://github.com/atlassian/openapi-request-validator/blob/master/docs/FEATURES.md), whereas swagger-request-validator-core only [supported OpenAPI v3.0](https://github.com/atlassian/openapi-request-validator/blob/swagger-request-validator-2.46.1/docs/FEATURES.md).

bbortt:
> Some thoughts on Java 21 in relation to this PR: Quarkus 4 is planned for November 2026 with Java 21 as the baseline, and Apache Camel will likely align with that as well. It might make sense for Citrus to follow the same approach. That would mean releasing Citrus 5.0.x with Java 17 for now, and then moving to Java 21 as the baseline toward the end of 2026.

PR #1542, opened 2026-05-28
fix(deps): replace org.apache.activemq with org.apache.artemis

bbortt:
migrate dependencies of group `org.apache.activemq` to group `org.apache.artemis`.


christophd (approved):
Thank you!

Issue #809, opened 2021-10-11
Bug: HTTP mocks can't handle complex query params that include commas


bthdimension:
**Citrus Version**
2.8.0, but the code causing the issue is still present in 3.0.0

**Expected behavior**
When calling a Citrus HTTP Server, I expect it to handle complex but correctly formated query parameters and continue to validate my headers and payloads as usual.

**Actual behavior**
Citrus can't handle the HTTP request and throws an error before getting to validate it. This is due to the following lines:

- https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-http/src/main/java/com/consol/citrus/http/controller/HttpMessageController.java#L163
- https://github.com/citrusframework/citrus/blob/main/endpoints/citrus-http/src/main/java/com/consol/citrus/http/message/HttpMessage.java#L218

Splitting by commas doesn't work if the values of the query params are more complex and contain commas. Why not just split by "&"?

**Test case sample**
Our query params look like this:
`?$top=3,$select=ComponentTypeName,%20Name%20,Description,$filter=ArchitectureName%20eq%20'Helsana%20Current%20State'%20and%20(Properties/any(p:p/Name%20eq%20'Inventar-ID'%20and%20(p/Value%20eq%20'17589'))),$expand=Properties($select=Name,%20Value;$filter=Name%20eq%20'Inventar-ID'%20or%20Name%20eq%20'App_Index'%20or%20Name%20eq%20'Status'%20or%20Name%20eq%20'Service%20Modell'%20or%20Name%20eq%20'Alias'%20or%20Name%20eq%20'Description'%20or%20Name%20eq%20'Serviceklasse'),%20inconnections($select=Name;$filter=Name%20eq%20'Service%20Owner'%20or%20Name%20eq%20'Applikationsowner'%20or%20Name%20eq%20'Benutzervertreter'%20or%20Name%20eq%20'Hersteller'%20or%20Name%20eq%20'Gesch%C3%A4ftsbereich';$expand=Properties($select=Name,Value;$filter=Name%20eq%20'Bemerkung'),SourceComponent($select=Name;$expand=Properties($select=Name,Value;$filter=Name%20eq%20'Benutzer-ID'))),$orderby=Name%20asc`

The error is:
`java.lang.IllegalStateException: Duplicate key Value;$filter (attempted merging values Name%20eq%20'Bemerkung') and Name%20eq%20'Benutzer-ID'))))`

novarx:
Still valid in 4.0.

[...]

PR #1569, opened 2026-06-22
fix(issue#809) spliting query param by &

otniel123:
Here I fixed the separation of query params by comma, instead, now it is separeted by ampersand (&).


christophd:
Many thanks for your contribution! I think you need to adjust the tests, too because the logic has changed. I would also love to see some new unit test that reproduces the initial error reported in #809

Apart from that I have one question. Your fix solves the commas being part of the query params, right. What about additional `&` characters in the new solution? Would that just hit the same error as before but now with `&` characters in the query params? I guess we would need to use some URL escape/unescape logic so users can use encoded strings for reserved characters such as `,` or `&`

christophd:
Issue has been solved with #1641 

Many thanks for your work!

PR #1605, opened 2026-07-14
fix(#1582): Add receive wait log output

### LINKED ISSUE #1582: Add log output when receive actions start waiting for messages
## Description

Blocking receive actions (e.g., waiting for an HTTP request on a server endpoint) produce no log output when they start waiting. If the receive eventually times out, the only indication is the timeout error — there is no earlier log line showing what the test was blocked on or when it started waiting.

## Current Behavior

No log output between starting a receive action and the timeout error:

```
<silence for 2 minutes>

TEST FAILED ... Action timeout after 120000 milliseconds. Failed to receive message on endpoint: 'jiraServer.inbound'
```

## Expected Behavior

Receive actions should log when they start waiting, e.g.:

```
Waiting to receive message on endpoint: 'jiraServer.inbound' (timeout: 120000ms)
```

This makes it immediately clear what the test is blocked on, especially when combined with other log lines from preceding actions.

## Additional Context

Reported in [apache/camel-kamelets#2867](https://github.com/apache/camel-kamelets/issues/2867). The lack of logging made it difficult to diagnose where the test was stuck.

w3lld1:
## Summary

- log the endpoint name and effective timeout at INFO immediately before a receive action starts waiting
- cover both regular and selective consumers while reusing the already-created selective consumer (avoids creating a second potentially stateful consumer)
- add focused regression tests for endpoint-default and action-specific timeouts
- declare the test's direct `log4j-core` dependency explicitly

## Validation

- `./mvnw -pl core/citrus-base -am -Dtest=ReceiveMessageActionLoggingTest -Dsurefire.failIfNoSpecifiedTests=false test` (2 tests passed)
- `./mvnw -pl core/citrus-base -am test` (11,424 tests passed across `citrus-api` and `citrus-base`)
- `git diff --check`

Fixes #1582

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
@@ -209,8 +209,17 @@ public class ReceiveMessageAction extends AbstractTestAction implements MessageA
      */
     private Message receive(TestContext context) {
         Endpoint messageEndpoint = getOrCreateEndpoint(context);
-        return receiveTimeout > 0 ? messageEndpoint.createConsumer().receive(context, receiveTimeout) :
-                messageEndpoint.createConsumer().receive(context, messageEndpoint.getEndpointConfiguration().getTimeout());
+        long timeout = getReceiveTimeout(messageEndpoint);
+        logWaiting(messageEndpoint, timeout);
+        return messageEndpoint.createConsumer().receive(context, timeout);
+    }
+
+    private long getReceiveTimeout(Endpoint messageEndpoint) {
+        return receiveTimeout > 0 ? receiveTimeout : messageEndpoint.getEndpointConfiguration().getTimeout();
+    }
+
+    private void logWaiting(Endpoint messageEndpoint, long timeout) {
+        logger.info("Waiting to receive message on endpoint: '{}' (timeout: {}ms)", messageEndpoint.getName(), timeout);
     }
 
     /**
@@ -226,15 +235,10 @@ public class ReceiveMessageAction extends AbstractTestAction implements MessageA
         Endpoint messageEndpoint = getOrCreateEndpoint(context);
         Consumer consumer = messageEndpoint.createConsumer();
         if (consumer instanceof SelectiveConsumer) {
-            if (receiveTimeout > 0) {
-                return ((SelectiveConsumer) messageEndpoint.createConsumer()).receive(
-                        context.replaceDynamicContentInString(selectorString),
-                        context, receiveTimeout);
-            } else {
-                return ((SelectiveConsumer) messageEndpoint.createConsumer()).receive(
-                        context.replaceDynamicContentInString(selectorString),
-                        context, messageEndpoint.getEndpointConfiguration().getTimeout());
-            }
+            long timeout = getReceiveTimeout(messageEndpoint);
+            logWaiting(messageEndpoint, timeout);
+            return ((SelectiveConsumer) consumer).receive(
+                    context.replaceDynamicContentInString(selectorString), context, timeout);
         } else {
             logger.warn("Unable to receive selective with consumer implementation: '{}'", consumer.getClass());
             return receive(context);
--- a/core/citrus-base/src/test/java/org/citrusframework/actions/ReceiveMessageActionLoggingTest.java
+++ b/core/citrus-base/src/test/java/org/citrusframework/actions/ReceiveMessageActionLoggingTest.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


core/citrus-base/src/test/java/org/citrusframework/actions/ReceiveMessageActionLoggingTest.java
```diff
@@ -0,0 +1,133 @@
...
+package org.citrusframework.actions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.Logger;
+import org.apache.logging.log4j.core.appender.AbstractAppender;
```
christophd, line 26: (Medium) The test directly depends on `log4j-core` classes (`AbstractAppender`, `Logger`, `LogEvent`, `Property`, `PatternLayout`), but `core/citrus-base/pom.xml` does not declare `log4j-core` as a test dependency. It works today through transitive resolution, but this is fragile. Consider adding an explicit dependency:

```suggestion
import org.apache.logging.log4j.core.LogEvent;
```

(The import itself is fine — the suggestion is to add `log4j-core` with `<scope>test</scope>` to `core/citrus-base/pom.xml` so this dependency is explicit rather than transitive.)

core/citrus-base/src/main/java/org/citrusframework/actions/ReceiveMessageAction.java
```diff
@@ -226,15 +235,10 @@ private Message receiveSelected(TestContext context, String selectorString) {
...
-                        context.replaceDynamicContentInString(selectorString),
-                        context, receiveTimeout);
-            } else {
-                return ((SelectiveConsumer) messageEndpoint.createConsumer()).receive(
-                        context.replaceDynamicContentInString(selectorString),
-                        context, messageEndpoint.getEndpointConfiguration().getTimeout());
-            }
+            long timeout = getReceiveTimeout(messageEndpoint);
+            logWaiting(messageEndpoint, timeout);
+            return ((SelectiveConsumer) consumer).receive(
```
christophd, line 240: Nice fix here — the old code called `messageEndpoint.createConsumer()` a second time instead of reusing the `consumer` variable from line 237. This could have caused issues with stateful consumer factories. Worth mentioning in the PR description as a bonus bug fix.

christophd (commented):
## Review: fix(#1582): Add receive wait log output

Nice work — this is a clean, well-scoped change that directly addresses the issue. The log message format matches what was requested, INFO is the right level for visibility, and the `getReceiveTimeout()` extraction eliminates duplication nicely.

### Confirmed Issues

**1. (Medium) Missing `log4j-core` test dependency declaration**
`ReceiveMessageActionLoggingTest` directly imports `log4j-core` classes (`AbstractAppender`, `Logger` cast, `LogEvent`, `Property`, `PatternLayout`) but `core/citrus-base/pom.xml` does not declare `log4j-core` as a test dependency. It compiles today through transitive resolution, but this is fragile — a dependency tree change could break the test without touching it. Adding an explicit `<dependency>` for `log4j-core` with `<scope>test</scope>` (version managed in root pom) would harden this:

```xml
<!-- Test scoped dependencies -->
<dependency>
    <groupId>org.apache.logging.log4j</groupId>
    <artifactId>log4j-core</artifactId>
    <scope>test</scope>
</dependency>
```

### Observations

**2. PR also fixes double `createConsumer()` call**
The old `receiveSelected()` called `messageEndpoint.createConsumer()` twice — once on line 227 to check `instanceof SelectiveConsumer`, then again on lines 230/234 to actually call `.receive()`. The PR correctly reuses the `consumer` variable. This is a nice bonus fix (double consumer creation could cause issues with stateful consumer factories) — worth calling out in the PR description for future `git blame` readers.

**3. AssertJ usage is consistent with the project**
The test uses AssertJ, which is available via the root pom's global `<dependencies>` block. While other tests in the `actions/` subdirectory use TestNG assertions, AssertJ is used in ~13 other test files across `citrus-base`. No issue here.

---

**Note:** This review covers project rules, conventions, and code correctness visible from the diff. It does not replace specialized review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

_This review was generated by AI (Claude) on behalf of Christoph Deppisch and may contain inaccuracies. Please verify all suggestions before applying._

w3lld1:
Thanks for the review. I pushed `bfa6ae043` to declare `log4j-core` explicitly as a test-scoped dependency, and updated the PR description to call out the consumer reuse.

Fresh validation:
- `./mvnw -pl core/citrus-base -am -Dtest=ReceiveMessageActionLoggingTest -Dsurefire.failIfNoSpecifiedTests=false test` — 2 passed
- `./mvnw -pl core/citrus-base -am test` — 11,424 passed across `citrus-api` and `citrus-base`
- `git diff --check`

christophd (approved):

PR #1623, opened 2026-07-23
delegate name and actor to anonymous inner class

### LINKED ISSUE #1622: `DefaultTestActionBuilder` produces unnamed `TestAction` instances
## Description

When using `DefaultTestActionBuilder.action(...)` (e.g., for final actions), the built `TestAction` may have no name (`getName()` is null/empty), even when the original delegate action has a meaningful class/name context.

A simple example looks like this:

```java
@Test
@CitrusTest
void addAndExecuteFinalAction(@CitrusResource TestContext testContext) {
    assertThat(testContext)
        .isNotNull();

    testContext.doFinally(action(_ -> FINAL_ACTION_HAS_BEEN_EXECUTED.set(true)));
}
```

This causes downstream tooling/listeners that record actions by name to fail or behave unpredictable.

## Affected Version

Observed with `org.citrusframework:citrus-base:5.0.0-M2`.

## Minimal Reproduction
1. Create a test using Citrus and register a final action via:
   `testContext.doFinally(action(_ -> ...))`
2. Attach a listener/collector that calls `testAction.getName()` for each executed action and uses it as a map key.

```java
public class MyTestActionListener implements TestActionListener {

    @Override
    public void onTestActionStart(TestCase testCase, TestAction testAction) {
        System.out.println(testAction.getName()); // sometimes is `null`
    }
}
```

3. Execute the test.

## Actual Behavior

- The action received during execution may have:
  - `getName() == null` (or blank), and/or
  - runtime class name like `org.citrusframework.DefaultTestActionBuilder$1` (anonymous wrapper class)
- This can trigger runtime failures in consumers that require non-null action names (e.g., `ConcurrentHashMap.computeIfAbsent` with null key).

The cause is [this anonymous class inside the `DefaultTestActionBuilder`](https://github.com/citrusframework/citrus/blob/main/core/citrus-base/src/main/java/org/citrusframework/DefaultTestActionBuilder.java#L42).

## Expected Behavior

Built actions should expose a stable, non-empty logical name so listeners/instrumentation can safely identify and aggregate action executions.

## Impact

- Action execution can fail with `NullPointerException`.
- Action labels become inconsistent/non-semantic (`DefaultTestActionBuilder$1`) when consumers fallback to class names.
- Harder observability and debugging for executed action sequences.

bbortt:
closes https://github.com/citrusframework/citrus/issues/1622.

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/DefaultTestActionBuilder.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/DefaultTestActionBuilder.java
@@ -19,6 +19,9 @@ package org.citrusframework;
 import org.citrusframework.actions.AbstractTestAction;
 import org.citrusframework.context.TestContext;
 
+import static java.util.Objects.nonNull;
+import static org.citrusframework.util.StringUtils.hasText;
+
 public class DefaultTestActionBuilder extends AbstractTestActionBuilder<AbstractTestAction, AbstractTestActionBuilder<AbstractTestAction, ?>> {
 
     private final TestAction delegate;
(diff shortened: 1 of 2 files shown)


christophd (approved):
lgtm, thanks

Issue #1678, opened 2026-08-06
Remote server with TestNG Engine: defining packages in run configuration is not working as expected


cquoss:
**Citrus Version**
4.10.3

**Expected behavior**
When i add packages to the run configuration of the citrus remote maven plugin to search for tests there the observed behaviour is not as expected.
The tests defined in these packages should be found as test candidates and executed.

**Actual behavior**
The tests are not found and therefore also not executed.

**Test case sample**
No test available in github as of now, but i can directly link the part of the code here that is responsible for this behaviour:

https://github.com/citrusframework/citrus/blob/c8ef28b415145dea95d8a5859e45c27ecd6bcd9a/core/citrus-base/src/main/java/org/citrusframework/main/scan/JarFileTestScanner.java#L59

Line that reads now like this ...

`boolean isTestClass = (packageIsEmpty || packageAsPath.startsWith(entry.getName()))` 

... should read like this ...

`boolean isTestClass = (packageIsEmpty || entry.getName().startsWith(packageAsPath))` 

This has been successfully tested in our pipelines at Union Investment with a patched version of citrus-base-4.10.3.

I will try to provide a MR for this.

cquoss:
I'll give up here. Maybe a little too quick but i cannot figure it out. The guideline states i should open the PR against the v2.7-bugfix branch. But this branch is outdated, the code is completely different from main.

I'm confused. Please advise.

christophd:
@cquoss thanks for providing the MR to help fixing the issue! 

It depends on what branch you are working on with your changes. In case you are working on top of main branch your PR should be targeting main branch, too.

You can also checkout 4.10.x branch and put your changes on top of this, you should then open the PR towards 4.10.x branch.

Either way I will take care of backporting this to main branch (Citrus 5.0) or 4.10.x once your PR is merged.

BTW which guide have you been following to use v2.7-bugfix branch?

cquoss:
@christophd OK, i now created the PR against main: #1680

Hope this is OK now.

cquoss:
Ah you asked about the guide, one moment: ah, i cannot reproduce. On creation of first PR there was a pop-up on the right side of the screen stating: 'This seems to be your first PR in this project. Please be sure to follow the guidelines'

Sth like that. And there was this link to the documenation stating bug fixes should be targeted at the v2.7-bugfix branch.

cquoss:
[...]

PR #1679, opened 2026-08-06
Fix/1678:  jar-test-file-scanner - Fix test candidate detection in provided packages

cquoss:
See issue #1678 

This is a simple fix. The start of the jar entry name needs to be compared to the package path to decide if the jar entry is a potential test candidate, not the other way round.


cquoss:
I will open a new PR against the main branch. This is fruitless,

Issue #1682, opened 2026-08-07
JarFile Test Scanner should not inspect META-INF entries


cquoss:
**Citrus Version**
4.10.3

**Expected behavior**
When running the citrus tests from a fat uber-jar created by the citrus remote maven plugin and having not narrowed down the packages to scan in the run configuration and having the citrus-ftp dependency in the classpath without exclusions all test execution works fine.

**Actual behavior**
The test logs state that a class file named ...

`META-INF/versions/25/org/bouncycastle/jcajce/provider/kdf/pbkdf2/PBKDF2Spi$PBKDF2with8BIT.class`

... was detected as a test candidate class. After that the citrus remote server stops working. The GET /results?timeout=15000 requests from the citrus remote maven plugin running in our gitlab pipeline are being answered immediately by the remote server (probably with some http 500 respone, not investigated further). But the remote plugin also not handles the response well but immediatly sends the next GET request, resulting in log spilling.

**Test case sample**
Again, no test case, sorry, just the observation above and the code here:

https://github.com/citrusframework/citrus/blob/89c34bed310c9c28fc7e441e071777b5a910f3e2/core/citrus-api/src/main/java/org/citrusframework/TestClass.java#L63

The transformed className in this case looks like this:

`META-INF.versions.25.org.bouncycastle.jcajce.provider.kdf.pbkdf2.PBKDF2Spi` (not sure if the inner class path really is missing here, but that is not the point)

The point is that this results in a CNFE that should raise itself a citrus runtime exception. I do not know why we do not see this in our logs, thats sth for another issue.

But for this issue here i would suggest to simply exclude the META-INF folders from the test class search in the JarFile Test Scanner.  And add the description of this bahaviour also to the citrus documentation ("Do not put your test classes to run under the META-INF folder, they will not be detected there"). Sth like that.

PR #1684, opened 2026-08-07
fix(#1682): jar-file-test-scanner - jar entries from META-INF folders are no test candidates

cquoss:
Again a rather simple fix. Maybe documentation needs an update, too. Then please point me to it and i will also amend it.

The whole test class detection can be refactored to stream with filter predicate, but that should be done in a separate refactoring ticket.


christophd:
lgtm, thanks

Issue #1591, opened 2026-07-13
Flaky test: HttpParallelRequest_3_IT fails with ConcurrentModificationException


christophd:
## Description

`HttpParallelRequest_3_IT` (executed as `HttpServerIT.parallelRequestsIT`) is flaky in GitHub Actions. It fails intermittently with a `ConcurrentModificationException` caused by concurrent access to the `RestTemplate` interceptors list during parallel HTTP request sending.

## Steps to Reproduce

1. Run `HttpParallelRequest_3_IT` in CI (GitHub Actions) — the test sends multiple HTTP requests in parallel
2. The failure is intermittent and depends on thread scheduling / timing

## Expected Behavior

Parallel HTTP requests should be sent without thread-safety issues.

## Actual Behavior

The test fails with:

```
org.citrusframework.exceptions.CitrusRuntimeException: java.util.ConcurrentModificationException
    at org.citrusframework.actions.SendMessageAction.lambda$doExecute$1(SendMessageAction.java:165)
    ...
Caused by: java.util.ConcurrentModificationException
    at java.base/java.util.ArrayList.sort(ArrayList.java:1723)
    at org.springframework.core.annotation.AnnotationAwareOrderComparator.sort(AnnotationAwareOrderComparator.java:109)
    at org.springframework.http.client.support.InterceptingHttpAccessor.setInterceptors(InterceptingHttpAccessor.java:70)
    at org.citrusframework.http.client.HttpEndpointConfiguration.getRestTemplate(HttpEndpointConfiguration.java:240)
    at org.citrusframework.http.client.HttpClient.send(HttpClient.java:133)
    at org.citrusframework.actions.SendMessageAction.lambda$doExecute$1(SendMessageAction.java:160)
```

## Additional Context

Root cause: When multiple threads send HTTP requests in parallel using the same `HttpClient`, each call to `HttpClient.send()` invokes `HttpEndpointConfiguration.getRestTemplate()`, which calls `setInterceptors()` on the shared `RestTemplate`. Spring's `InterceptingHttpAccessor.setInterceptors()` sorts the interceptors list in place via `AnnotationAwareOrderComparator.sort()`. When two threads hit this concurrently, the underlying `ArrayList.sort()` throws `ConcurrentModificationException`.

This is not a test-only issue — any production use of parallel sends through a shared `HttpClient` could trigger the same race condition.

PR #1694, opened 2026-08-11
fix(#1690): make HttpEndpointConfiguration RestTemplate access thread-safe

### LINKED ISSUE #1690: Flaky test: HttpClientTest.testConcurrentGetRestTemplate fails with ConcurrentModificationException
## Description

The test `HttpClientTest.testConcurrentGetRestTemplate` is flaky in CI — it intermittently fails with a `ConcurrentModificationException` when 10 threads call `HttpEndpointConfiguration.getRestTemplate()` concurrently.

## Root Cause

`HttpEndpointConfiguration.getRestTemplate()` is not thread-safe. There are multiple race conditions:

1. **Check-then-act on `restTemplate`** — the `if (restTemplate == null)` guard and subsequent assignment is not synchronized, so multiple threads can enter the block simultaneously.
2. **Unsynchronized `clientInterceptors` list** — `clientInterceptors` is a plain `ArrayList`. When multiple threads concurrently execute `new ArrayList<>(clientInterceptors)` (which iterates the list) while Spring's `InterceptingHttpAccessor.setInterceptors()` sorts it in-place, a `ConcurrentModificationException` is thrown.
3. **`restTemplate` field is not volatile** — writes may not be visible to other threads.

## Stack Trace

```
java.util.ConcurrentModificationException
    at java.base/java.util.ArrayList.sort(ArrayList.java:1723)
    at org.springframework.core.annotation.AnnotationAwareOrderComparator.sort(AnnotationAwareOrderComparator.java:109)
    at org.springframework.http.client.support.InterceptingHttpAccessor.setInterceptors(InterceptingHttpAccessor.java:70)
    at org.citrusframework.http.client.HttpEndpointConfiguration.getRestTemplate(HttpEndpointConfiguration.java:240)
```

## Affected Code

- `endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpEndpointConfiguration.java` — `getRestTemplate()` (line ~237)
- `endpoints/citrus-http/src/test/java/org/citrusframework/http/client/HttpClientTest.java` — `testConcurrentGetRestTemplate()` (line ~528)

## Additional Context

The test was specifically written to verify thread-safe access to `getRestTemplate()`, but the production code does not yet provide that guarantee. A fix should synchronize access to the shared mutable state (e.g. `synchronized` block, `volatile` field, or defensive copying of the interceptor list).

kalayciburak:
## Summary

`HttpEndpointConfiguration.getRestTemplate()` lazily creates a shared `RestTemplate` without synchronization. Concurrent callers can race Spring's in-place interceptor sort and throw `ConcurrentModificationException`, which is what makes `HttpClientTest.testConcurrentGetRestTemplate` flake in CI.

This change:

- Marks `getRestTemplate`, `setRestTemplate`, and `setClientInterceptors` as `synchronized` so lazy init and interceptor updates share one monitor
- Strengthens the concurrent unit test (32 threads + `assertSame` on the shared instance)

Fixes #1690

## Test plan

- [x] `./mvnw -pl endpoints/citrus-http test -Dtest=HttpClientTest` — 20/20 GREEN

### DIFF
--- a/endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpEndpointConfiguration.java
+++ b/endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpEndpointConfiguration.java
@@ -152,7 +152,7 @@ public class HttpEndpointConfiguration extends AbstractPollableEndpointConfigura
      * Sets the restTemplate.
      * @param restTemplate the restTemplate to set
      */
-    public void setRestTemplate(RestTemplate restTemplate) {
+    public synchronized void setRestTemplate(RestTemplate restTemplate) {
         clientInterceptors.addAll(restTemplate.getInterceptors());
         restTemplate.setInterceptors(new ArrayList<>(clientInterceptors));
         this.restTemplate = restTemplate;
@@ -234,7 +234,7 @@ public class HttpEndpointConfiguration extends AbstractPollableEndpointConfigura
      * Gets the restTemplate.
      * @return the restTemplate
      */
-    public RestTemplate getRestTemplate() {
+    public synchronized RestTemplate getRestTemplate() {
         if (restTemplate == null) {
             restTemplate = new RestTemplate();
             restTemplate.setInterceptors(new ArrayList<>(clientInterceptors));
@@ -281,7 +281,7 @@ public class HttpEndpointConfiguration extends AbstractPollableEndpointConfigura
      * Sets the clientInterceptors on this implementation's rest template.
      * @param clientInterceptors the clientInterceptors to set
      */
-    public void setClientInterceptors(List<ClientHttpRequestInterceptor> clientInterceptors) {
+    public synchronized void setClientInterceptors(List<ClientHttpRequestInterceptor> clientInterceptors) {
         this.clientInterceptors = clientInterceptors;
 
         if (restTemplate == null) {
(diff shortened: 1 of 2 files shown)


christophd (approved):
Nice fix — correct root cause identified and addressed with a minimal, well-scoped change.

**What's good:**

- The prior fix (#1591) only made defensive copies of the interceptor list but left `getRestTemplate()` unsynchronized. This PR correctly adds `synchronized` to `getRestTemplate()`, `setRestTemplate()`, and `setClientInterceptors()` — the three methods that read/write the shared `restTemplate` and `clientInterceptors` fields. Since Java's `synchronized` is reentrant, the `setClientInterceptors()` → `getRestTemplate()` call path is safe.
- `HttpEndpointConfiguration` has no subclasses, so synchronizing on `this` introduces no risk of lock-ordering issues.
- Test strengthened with 32 threads and `assertSame` to verify all threads get the same instance.
- Branch naming, commit format, and PR body all follow project conventions.

**Minor (non-blocking):**

- The `first` variable in the test could be extracted before the loop (e.g. `RestTemplate first = futures.get(0).get()`) instead of the null-check pattern inside the loop — purely cosmetic.
- CI build is still pending at the time of this review — worth confirming it passes before merge.

LGTM — thanks for the contribution!

_This review was generated by an AI agent on behalf of Christoph Deppisch and may contain inaccuracies. Please verify all suggestions before applying._

PR #1696, opened 2026-08-11
fix(http): synchronize RestTemplate initialization

### LINKED ISSUE #1690: Flaky test: HttpClientTest.testConcurrentGetRestTemplate fails with ConcurrentModificationException

Anikesh348:
## Summary

- serialize access to the lazily initialized `RestTemplate`
- prevent concurrent initialization from mutating the interceptor list at the same time

Fixes #1690

## Testing

`./mvnw -pl endpoints/citrus-http -am -Dtest=HttpClientTest#testConcurrentGetRestTemplate -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpEndpointConfiguration.java
+++ b/endpoints/citrus-http/src/main/java/org/citrusframework/http/client/HttpEndpointConfiguration.java
@@ -234,7 +234,7 @@ public class HttpEndpointConfiguration extends AbstractPollableEndpointConfigura
      * Gets the restTemplate.
      * @return the restTemplate
      */
-    public RestTemplate getRestTemplate() {
+    public synchronized RestTemplate getRestTemplate() {
         if (restTemplate == null) {
             restTemplate = new RestTemplate();
             restTemplate.setInterceptors(new ArrayList<>(clientInterceptors));


christophd:
Hi @Anikesh348, thanks for taking the time to work on this! Just a heads-up — PR #1694 already addresses the same issue (#1690) with a very similar approach, and it has been approved for merge. We'll be going with that one to avoid duplicate changes.

That said, your analysis of the problem was spot-on, and your fix follows the right approach. Please don't be discouraged — we'd love to see you contribute again! There are more issues labeled `good first issue` if you're looking for something to pick up next.

Thanks again and welcome to the project!

PR #1704, opened 2026-08-12
fix(#1654): compile Java test sources into a temp output directory

### LINKED ISSUE #1654: citrus-junit-jupiter: Subsequent Maven builds without clean fail due to stale compiled test resources
## Description

The `citrus-junit-jupiter` module requires `mvn clean` before every `mvn verify`. Running `mvn verify` twice without `clean` fails with:

```
JavaTest (wrong name: org/citrusframework/junit/jupiter/integration/java/JavaTest)
```

## Root Cause

`JavaTestLoader.doLoad()` in `citrus-base` compiles Java test resource files **in-place** without a `-d` flag:

```java
int success = compiler.run(null, null, null, javaSource.file().getAbsolutePath());
```

This causes `javac` to write `.class` files next to the `.java` sources inside `target/test-classes/`. Since the resource files (e.g. `JavaTest.java`, `EchoTest.java`, `DelayTest.java`) have **no `package` declaration**, the compiled class internal name (e.g. `JavaTest`) does not match the filesystem path (`org/citrusframework/junit/jupiter/integration/java/JavaTest.class`).

On the first build this goes unnoticed because surefire hasn't scanned these classes yet. On a second `mvn verify` (without `clean`), surefire discovers the stale `.class` files and the JVM rejects them due to the name mismatch.

## Steps to Reproduce

```bash
cd runtime/citrus-junit-jupiter
mvn verify        # passes
mvn verify        # fails
mvn clean verify  # passes again
```

## Expected Behavior

`mvn verify` should succeed on repeated runs without requiring `mvn clean`.

## Suggested Fix

Compile into a temporary directory using the `-d` flag in `JavaTestLoader` so `.class` files don't pollute `target/test-classes`.

## Affected Files

- `core/citrus-base/src/main/java/org/citrusframework/common/JavaTestLoader.java` (line 57)
- `runtime/citrus-junit-jupiter/src/test/resources/org/citrusframework/junit/jupiter/integration/java/JavaTest.java`
- `runtime/citrus-junit-jupiter/src/test/resources/org/citrusframework/junit/jupiter/simple/EchoTest.java`
- `runtime/citrus-junit-jupiter/src/test/resources/org/citrusframework/junit/jupiter/simple/DelayTest.java`

kalayciburak:
## Summary

`JavaTestLoader` compiled on-disk Java test sources with `javac` without a `-d` output directory. For default-package sources living under a nested path in `target/test-classes`, that wrote `.class` files next to the `.java` resources. Surefire then discovered those stale classes on a subsequent `mvn verify` (without `clean`) and the JVM rejected them with wrong-name errors.

This change:

- Compiles into a temporary directory via `-d`
- Loads the compiled class from that output root
- Cleans up the temporary output after load
- Adds unit coverage for default-package and packaged sources

Fixes #1654

## Test plan

- [x] `./mvnw -pl core/citrus-base test -Dtest=JavaTestLoaderTest` — 2/2 GREEN
- [x] `./mvnw -pl runtime/citrus-junit-jupiter verify` twice without `clean` — no stale `JavaTest.class` / `EchoTest.class` / `DelayTest.class` beside sources; `JavaTestLoader_IT` 7/7 GREEN

### DIFF
--- a/core/citrus-base/src/main/java/org/citrusframework/common/JavaTestLoader.java
+++ b/core/citrus-base/src/main/java/org/citrusframework/common/JavaTestLoader.java
@@ -18,13 +18,14 @@ package org.citrusframework.common;
 
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
-import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
+import java.util.Comparator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import javax.tools.JavaCompiler;
 import javax.tools.ToolProvider;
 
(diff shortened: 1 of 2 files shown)


core/citrus-base/src/main/java/org/citrusframework/common/JavaTestLoader.java
```diff
@@ -132,24 +143,22 @@ private static String extractPackageName(Resource javaSource) throws IOException
...
+                try {
+                    Files.deleteIfExists(path);
+                } catch (IOException e) {
+                    // Best-effort cleanup of temporary compile output
+                }
+            });
+        } catch (IOException e) {
+            // Best-effort cleanup of temporary compile output
+        }
     }
```
christophd, line 163: Nit: Both `catch (IOException)` blocks here silently swallow exceptions. While best-effort cleanup is reasonable for temp directories, a `logger.debug` or `logger.trace` call would help diagnose cleanup failures in CI environments. Not blocking.

christophd (approved):
Nice work on this fix, @kalayciburak! The approach is clean and well-scoped. A few notes:

**Branch naming (minor):** Per project conventions, fix branches follow `fix/<ISSUE_NUMBER>` (i.e. `fix/1654`). The current branch `fix/1654-java-test-loader-compile-outdir` uses the feature branch pattern instead. No action needed here, just a note for future contributions.

**Positive highlights:**
- The `URLClassLoader` is now properly closed via try-with-resources (fixes a resource leak in the original code)
- The `sourcePackageName` rename eliminates field shadowing for clarity without changing behavior
- The `finally` block ensures temp directory cleanup even on failure
- Test coverage for both default-package and packaged source scenarios is thorough

LGTM — thanks for the contribution!

_This review was generated by an AI agent on behalf of Christoph Deppisch and may contain inaccuracies. Please verify all suggestions before applying._

PR #1715, opened 2026-08-24
feat(#1714): add Playwright connector for browser automation

### LINKED ISSUE #1714: Add Playwright connector for browser automation
### Motivation

Citrus ships Selenium support (`connectors/citrus-selenium`) but has no Playwright
equivalent. Teams standardizing on Playwright currently cannot drive modern browser flows
(contexts, pages, locators, tracing, network/console capture) natively from Citrus tests
and must fall back to custom glue code or non-Citrus tooling.

### Proposal

A new connector module `connectors/citrus-playwright` providing:

- Java DSL entry point `$(playwright()...)` via `PlaywrightActionBuilder`
- 30+ browser actions: navigation, input, keyboard/mouse, frames, dialogs,
  cookies and local/session storage, console & network capture, tracing,
  downloads, PDF export, page objects
- Flexible browser binding: explicit endpoint/name reference, ambient thread binding
  via `start()`/`stop()`, Citrus variable, or unique-bean resolution — always fail-fast
  with a clear resolution report when ambiguous
- No Spring dependency — only `citrus-api`, `citrus-base` and
  `com.microsoft.playwright:playwright` (same shape as `citrus-agent-connector`,
  Java-DSL-first)

### Documentation

Reference-manual page `src/manual/endpoint-playwright.adoc`, modeled on
`endpoint-selenium.adoc` and included in `src/manual/index.adoc`.

### Testing

Unit tests plus opt-in integration tests against real Chromium; ITs skip gracefully
when Playwright browser binaries are absent so normal builds stay hermetic.

### Contribution

I have this connector working in an internal project and will contribute it via a PR
referencing this issue, ported to upstream `master` conventions. The connector ships
XML/YAML config-DSL parity with `citrus-selenium` (declarative `playwright` action
namespace, endpoint annotation parser); feedback welcome if maintainers would rather
split the declarative DSL into a follow-up.

asuruceanu:
Resolves #1714

### Motivation

Citrus ships Selenium support (`connectors/citrus-selenium`) but has no Playwright equivalent. Teams standardizing on Playwright currently cannot drive modern browser flows (contexts, pages, locators, tracing, network/console capture) natively from Citrus tests and must fall back to custom glue code or non-Citrus tooling.

### What this adds

A new connector module `connectors/citrus-playwright` providing:

- Java DSL entry point `$(playwright()...)` via `PlaywrightActionBuilder`
- Browser actions: navigation, input, keyboard/mouse, wait conditions, screenshots, JavaScript evaluation, structured `verify`/`extract` with sanitized failure messages
- Flexible browser binding: explicit endpoint/name reference, ambient thread binding via `start()`/`stop()`, Citrus variable, or unique-bean resolution — always fail-fast with a clear resolution report when ambiguous
- XML/YAML config DSL parity with `citrus-selenium` (declarative `playwright` action namespace) plus a `@PlaywrightBrowserConfig` endpoint annotation parser
- No Spring dependency — only `citrus-api`, `citrus-base`, `jakarta.xml.bind-api` and `com.microsoft.playwright:playwright` (driver version managed in root POM next to `selenium.version`)
- Reference manual page `src/manual/endpoint-playwright.adoc`, modeled on the Selenium page

### Design decisions

- **Ambient thread binding**: starting a browser binds it to the executing thread; actions resolve browser by explicit binding → thread scope → variable → unique bean. Multi-browser tests bind chains explicitly; cross-thread sharing fails fast.
- **Observability**: endpoints keep bounded console/network buffers and can persist failure evidence (screenshot, page source, trace, logs, summary) with secret redaction applied.
- **Declarative DSL**: hand-written JAXB/snakeyaml wrapper packages (`xml`/`yaml`) registered through `META-INF/citrus/{xml,yaml}/builder/playwright`; no Spring XML namespace tooling and no XSD codegen — modern DSL only.

### API adaptations vs internal baseline (drift log)

The connector was developed against Citrus 4.9.3 APIs and ported to current `master`. File-by-file:

| File | Drift | Fix |
|---|---|---|
| `endpoint/PlaywrightBrowser.java` | `ShutdownPhase` moved to `org.citrusframework.api.common` | import updated |
| `endpoint/PlaywrightBrowser.java` | `Thread.threadId()` requires Java 19+, upstream targets 17 | replaced with `Thread.currentThread().getId()` |
| `endpoint/PlaywrightEndpointBuilder.java` | `AbstractEndpointBuilder` moved to `org.citrusframework.base.endpoint` | import updated |
| `config/annotation/PlaywrightBrowserConfigParser.java` | `AnnotationConfigParser#parse` gained a `TestContext` parameter | signature updated; dynamic content replacement restored for string attributes |
| `yaml/*.java` | `@SchemaProperty` moved to `org.citrusframework.api.yaml` (with new `module()` attribute) | imports updated |
| test sources | converted from JUnit 5 to TestNG to match existing connector conventions (`AbstractTestNGUnitTest` scaffolding as in `citrus-agent-connector`) | mechanical conversion, same assertions |

### Testing

- 243 unit tests: per-action coverage for each browser action, plus YAML/XML loader tests that load **and execute** complete declarative test cases against a proxy-stubbed browser (no real browser needed)
- Two test doubles keep the unit suite browser-free: `StubStartedBrowser` (dynamic proxies, used for lifecycle and browser-resolution tests) and `MockPlaywrightBrowser` (Mockito mocks for the Playwright driver objects, so interaction-heavy actions can be exercised and verified)
- `PlaywrightBrowserIT` drives real Chromium against local HTML fixtures (interaction, verification, extraction, dialogs/downloads); it skips gracefully via `SkipException` when Chromium is unavailable, so normal builds stay hermetic
- `mvn test -pl connectors/citrus-playwright` → 243/243 green; `mvn verify -pl connectors/citrus-playwright` runs the ITs

Happy to adjust naming/structure per maintainer feedback. If declarative DSL parity is better split into a follow-up PR, the `xml`/`yaml` packages + builder resources are self-contained and easy to separate.


connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/yaml/Playwright.java
```diff
@@ -0,0 +1,203 @@
...
+        this.browserName = browser;
+    }
+
+    @SchemaProperty(kind = ACTION, group = PLAYWRIGHT_GROUP)
+    public void setStart(Start builder) {
+        this.builder = builder;
+    }
+
+    @SchemaProperty(kind = ACTION, group = PLAYWRIGHT_GROUP)
+    public void setStop(Stop builder) {
```
christophd, line 75: The `module` attribute is missing from the `@SchemaProperty` annotations on action setters. The Selenium module sets `module = SELENIUM_MODULE` on every `@SchemaProperty(kind = ACTION, ...)` annotation — this metadata is used by the JSON schema generator.

Consider defining a constant and adding it:

```java
private static final String PLAYWRIGHT_MODULE = "citrus-playwright";
```

and then:

```suggestion
    @SchemaProperty(kind = ACTION, group = PLAYWRIGHT_GROUP, module = PLAYWRIGHT_MODULE)
```

connectors/citrus-playwright/scripts/install-playwright-runtimes.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
christophd: This script is Unix-only (bash) and duplicates what the `playwright-runtimes` Maven profile already provides cross-platform. Consider removing the entire `scripts/` directory — users can use `mvn verify -P playwright-runtimes -pl connectors/citrus-playwright` for both CI and local setup.

Additionally, if kept, there are two bugs:
- Line 5: `REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"` resolves to `connectors/`, not the repo root (needs `../../..`)
- Line 17: `-pl citrus-playwright` should be `-pl connectors/citrus-playwright` or `-pl :citrus-playwright`

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/actions/ScreenshotAction.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    protected void execute(PlaywrightBrowser browser, TestContext context) {
+        String resolvedPath = LocatorResolver.resolve(path, context);
+        byte[] bytes = browser.getCurrentPage().screenshot(new com.microsoft.playwright.Page.ScreenshotOptions()
+                .setPath(Path.of(resolvedPath)));
+        if (variable != null) {
+            context.setVariable(variable, resolvedPath != null ? resolvedPath : bytes);
+        }
```
christophd, line 51: Minor: the `bytes` branch in this ternary is unreachable — `resolvedPath` is always non-null because `LocatorResolver.resolve(path, context)` returns a non-null string when `path` is non-null, and `path` defaults to `"target/playwright/screenshot.png"` in the builder.

If the intent was to store raw bytes when no path is explicitly set, the condition should check the original `path` field. Otherwise, simplify to just `resolvedPath`:

```suggestion
            context.setVariable(variable, resolvedPath);
```

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/endpoint/PlaywrightBrowser.java
```diff
@@ -0,0 +1,765 @@
...
+     * Playwright resources, and unregisters the endpoint from failure evidence.
+     */
+    public synchronized void stop() {
+        ActivePlaywrightBrowsers.unregister(this);
+        if (currentContext != null && getEndpointConfiguration().isTracingEnabled()) {
+            try {
+                Path tracePath = Path.of("target", "playwright", "trace.zip");
+                Files.createDirectories(tracePath.getParent());
+                currentContext.tracing().stop(new Tracing.StopOptions().setPath(tracePath));
+            } catch (Exception ignored) {
```
christophd, line 200: Nit: when both `tracingEnabled` and `captureFailureTrace` are true and a test fails, `context.tracing().stop()` may be called twice — once by `FailureEvidenceWriter` during the failure callback and again here during `stop()`. The second call is caught and swallowed (line 180), so it doesn't break anything, but the trace file written here may be empty. Consider guarding with a flag or checking tracing state before stopping.

christophd (changes requested):
## Review: feat(#1714): add Playwright connector for browser automation

Thank you for this impressive contribution — the module is well-researched, aligns closely with Citrus conventions, and the thread-safety model (`PlaywrightBrowserScope` with `WeakReference<TestContext>`, `assertActionThread()`) is genuinely well-designed. The effort to port from your internal project and align with upstream APIs is clearly evident throughout.

A few items that should be addressed before merging:

### Confirmed Issues

1. **Missing `META-INF/LICENSE.txt` and `META-INF/NOTICE.txt`** — All existing connector modules (`citrus-selenium`, `citrus-kubernetes`, `citrus-docker`, `citrus-sql`, `citrus-testcontainers`) include these files in `src/main/resources/META-INF/`. The new module should add them for Apache 2.0 compliance in the packaged JAR.

2. **Missing `module` attribute on `@SchemaProperty`** — See inline comment on `yaml/Playwright.java`.

3. **Remove `scripts/install-playwright-runtimes.sh`** — The script is Unix-only and redundant with the `playwright-runtimes` Maven profile, which works cross-platform. The Maven profile is sufficient for both CI automation and local environment preparation.

4. **No CI coverage for Playwright integration tests** — The ITs currently skip silently in the GitHub Actions workflow because the `playwright-runtimes` profile is not activated. Consider adding a dedicated CI job:
   ```yaml
   playwright-tests:
     runs-on: ubuntu-latest
     needs: build
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-java@v5
         with: { java-version: '17', distribution: 'temurin' }
       - run: ./mvnw verify -pl connectors/citrus-playwright -am -P playwright-runtimes
   ```

### Questions

- **`ReferenceResolverAware` propagation in XML `build()`**: Selenium's `xml/Selenium.java` propagates the reference resolver to sub-action builders. Playwright's skips this. Currently harmless (no sub-actions implement it), but deviates from the pattern — intentional?
- **Endpoint builder registration**: Selenium registers `browser=SeleniumBrowserBuilder` in `endpoint/builder/selenium`. Playwright only has `type=PlaywrightEndpointBuilder`. Intentional consolidation, or should a `browser=` alias be added?

### What's great

- API imports all match current `main` branch (`ShutdownPhase`, both `AbstractEndpointBuilder` variants, `@SchemaProperty`, `AnnotationConfigParser.parse(TestContext)`)
- META-INF SPI registrations exactly mirror the Selenium pattern
- POM registrations complete (module + managed dependency)
- Test naming conventions followed (`Test.java` / `IT.java`)
- Commit messages follow project conventions
- Documentation properly added and registered in `index.adoc`
- License headers on all 134 Java files
- Resource cleanup is robust with `*Quietly` close methods
- Browser resolution diagnostic reporting in `AbstractPlaywrightAction.resolveBrowser()` is excellent DX

**Note:** This review covers project rules, conventions, and structural correctness. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analysis (SonarCloud).

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/yaml/Playwright.java
asuruceanu: Done in d3f8b49 — added `private static final String PLAYWRIGHT_MODULE = "citrus-playwright";` and applied `module = PLAYWRIGHT_MODULE` to all 22 `kind = ACTION` setters.

One judgement call in the same commit, please push back if you'd rather keep it out: `PlaywrightEndpointBuilder` also carried no `@SchemaType(module = ...)`, which every other registered endpoint builder in the repo has (`SeleniumBrowserBuilder`, `DockerClientBuilder`, `KubernetesClientBuilder`, and the endpoint modules). It looked like the same schema-generator metadata gap, so I added it there too. I deliberately did not add `@XmlType`, since the Playwright endpoint is configured through `@PlaywrightBrowserConfig` rather than XML endpoint config.

connectors/citrus-playwright/scripts/install-playwright-runtimes.sh
asuruceanu: Removed the whole `scripts/` directory in 2abe50d.

Both bugs were real — `${SCRIPT_DIR}/../..` resolved to `connectors/`, and `-pl citrus-playwright` was a leftover from the pre-port layout — but I agree the Maven profile makes the script redundant, so it is gone rather than fixed. The reference manual already documents `mvn generate-test-resources -pl connectors/citrus-playwright -Pplaywright-runtimes`, so no docs change was needed.

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/actions/ScreenshotAction.java
asuruceanu: Confirmed and fixed in 4349147.

It is actually a little stronger than described: `Path.of(resolvedPath)` dereferences the value two lines above the ternary, so a null `resolvedPath` would already have thrown an NPE before the `bytes` branch could be reached. Storing the path is the documented behaviour, so I removed the dead branch and the now-unused `bytes` local rather than repairing the condition.

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/endpoint/PlaywrightBrowser.java
asuruceanu: Confirmed and fixed in 210ce10.

Tracing now stops through a guarded `stopTracing(Path)` on the endpoint. `FailureEvidenceWriter` and `stop()` both route through it, so the first caller wins and later calls are no-ops; the flag resets in `start()` so a restarted browser traces again. That keeps the failure trace from being overwritten by the empty archive teardown used to write.

Covered by `PlaywrightBrowserLifecycleTest.shouldStopTracingOnlyOnce`. I checked it fails with the guard removed (`Wanted 1 time ... But was 2 times`), so it is not a vacuous test.

asuruceanu:
Thanks for the detailed review — much appreciated. All four confirmed issues and both inline nits are addressed in 8 commits on top of the branch, one commit per item, and I have answered both questions by aligning with the Selenium/connector pattern. `mvn clean test -pl connectors/citrus-playwright` is 244/244 green.

### Confirmed issues

1. **`META-INF/LICENSE.txt` / `NOTICE.txt`** — added in b0d8f71, copied verbatim from `citrus-selenium` (all nine connectors ship byte-identical files).
2. **`module` on `@SchemaProperty`** — d3f8b49, see the inline reply. Also added `@SchemaType(module = "citrus-playwright")` to `PlaywrightEndpointBuilder`, which was the only registered endpoint builder in the repo without one.
3. **`scripts/install-playwright-runtimes.sh`** — removed in 2abe50d, see the inline reply.
4. **CI coverage** — 0e03800 adds a `playwright-tests` job to `.github/workflows/build.yml`, essentially your snippet plus the Maven cache step the existing `build` job uses:

```
./mvnw --no-transfer-progress verify -pl connectors/citrus-playwright -am -Pplaywright-runtimes
```

### Questions

- **`ReferenceResolverAware` propagation in the XML `build()`** — not intentional, just an omission. Added in 4317fbd, mirroring `xml/Selenium.java` for both the container-builder loop and the builder itself. As you say it is inert today, so this is purely about not deviating from the pattern.
- **Endpoint builder registration** — added `browser=` in 73f475c. Worth noting the two modules are shaped differently: `SeleniumEndpointBuilder` is a dispatcher that wraps `SeleniumBrowserBuilder`, whereas `PlaywrightEndpointBuilder` *is* the browser builder, so both keys point at the same class. The alias was still wrong to omit — `PlaywrightEndpoints.browser()` already existed in the Java DSL, so the SPI resource simply did not match the facade. If you would prefer the dispatcher shape instead for consistency with Selenium, I am happy to restructure.

### Not addressed

Nothing from the review is outstanding. Two of the fixes were judgement calls that go slightly beyond what was asked — the `@SchemaType` addition and the `browser=` alias pointing at the existing builder rather than a new dispatcher — so flagging them explicitly in case you want either changed.

christophd:
many thanks for addressing the review comments! 

regarding the new `playwright-tests` CI job. Are there any integration tests that really need the installed playwright browser? I know the review comments stated that tests are skipped silently, but is this really the case?

It seems like the Maven profile `playwright-runtimes` performs the installation but the tests would also run perfectly without it, right? Are there any integration tests planned that use the installed browsers? 

I just wonder if we should run this new job with every commit or with the nightly job only (to verify the runtime installation profile works).

asuruceanu:
Thanks! I checked this properly. Short version: yes, the browser is needed; no, the tests were not skipping; and the new job was not running any test at all.

**1. Do any ITs really need the installed browser?**

Yes. `PlaywrightBrowserIT` is the only IT in the module, and all 6 tests drive real Chromium against local HTML fixtures. The other 246 tests in the module are unit tests with mocks. So this class is the only end-to-end coverage of the endpoint. The tests exist today, they are not planned work.

**2. Were the tests skipped silently?**

No, that claim was wrong - it was mine, in the manual, and it is fixed now. Playwright's Java bindings download browsers themselves on the first `Playwright.create()` call, so on a clean runner the tests run. They only skip if that download is blocked.

**3. Does the profile only install, with the tests running fine without it?**

Not quite: the profile was stopping the tests from running. `com.microsoft.playwright.CLI.main` ends with `System.exit()`. The profile called it through `exec:java`, which runs in the Maven JVM, and since JDK 17 that exit can no longer be intercepted. Maven therefore stopped right after the install with exit code 0, before failsafe. The `playwright-tests` job installed Chromium and went green having run 0 tests.

**What `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` means in build.yml**

It is a Playwright environment variable that turns off Playwright's own automatic browser download.

Without it, the first `Playwright.create()` call installs every browser Playwright knows about - Chromium, Firefox and WebKit. It does this even when the profile has already installed Chromium, because it checks the whole set and fetches whatever is missing. That download runs inside the test.

With it, Playwright never installs anything by itself. The `playwright-runtimes` profile becomes the only thing that installs a browser, and it installs Chromium only. If the browser is missing, the tests skip instead of quietly downloading three browsers.

So both are needed: the profile installs Chromium, the environment variable stops everything else from being installed.

Measured with `verify -pl connectors/citrus-playwright` on an empty browser cache:

| config | tests | downloaded | time |
| --- | --- | --- | --- |
| neither | 6 run | all browsers | 5:45 |
| profile only | 6 run | all browsers | 4:29 |
| env var only | 6 skipped | nothing | 8s |
| both | 6 run | Chromium only | 1:36 |

With a warm cache and both: 13s.

**What I changed**

- `connectors/citrus-playwright/pom.xml` - the profile now uses `exec:exec` with a forked JVM, so the CLI's `System.exit()` no longer kills the build.
- `.github/workflows/build.yml` - removed the `playwright-tests` job; the `build` job sets `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1`, caches `~/.cache/ms-playwright`, and runs `install -Pplaywright-runtimes`.
- `PlaywrightBrowserIT` - moved the `chromiumAvailable()` probe into `@BeforeClass`; it used to launch a browser per test method. Class time drops from 10.0s to 5.3s. The skip line now reports 8 instead of 6, because TestNG counts the two configuration methods.
- `src/manual/endpoint-playwright.adoc` - fixed the incorrect skip/hermetic claim and documented the environment variable.

So there is no new job on every commit: the browser tests run in the existing `build` job, Chromium only, cached between runs. Worth knowing that the root build already ran this IT before this PR changed anything (failsafe is bound in the root pom and `skip.integration.tests` defaults to `${skipTests}`), so every PR was already downloading all browsers implicitly. That cannot happen now.

**If you would rather keep PR builds browser-free**

Only the profile moves. The environment variable stays where it is:

- `build` job: keep `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1`, remove `-Pplaywright-runtimes`. The tests then skip, nothing is downloaded, ~8s. If the environment variable were dropped as well, the job would go back to downloading all browsers and running the tests (5:45), which is what this PR did before these commits.
- new nightly job: needs both - the environment variable plus `-Pplaywright-runtimes` - and the `~/.cache/ms-playwright` cache step. The tests run, against Chromium only.

So the environment variable belongs in every job that builds this module, and the profile is what decides whether browsers are actually installed there.

I would keep it on every commit, since these are the only tests that would catch a regression in the endpoint and they cost about 13s warm. Happy to move it to nightly if you prefer.

christophd (approved):
LGTM now many thanks for this contribution!

Issue #1721, opened 2026-09-08
Array query parameters with Citrus variable expressions are not resolved in generated REST API test actions


tschlat:
**Citrus Version**
5.0.1

**Expected behavior**
Variable expressions in array-type query parameters are resolved against the test context, exactly as they are for `String` and `List` parameters. The outgoing request contains the resolved values.

**Actual behavior**
The unresolved expressions are used verbatim when building the query string, and the request fails.

bbortt:
closed in https://github.com/citrusframework/citrus/pull/1722.

PR #1722, opened 2026-09-08
fix(#1721): properly resolve query array parameters from citrus context

tschlat:
Fixed issue and added test in java and xml.


bbortt (approved):
thank you 🚀

PR #1728, opened 2026-09-16
fix/#1727: fix type of cause field in remote-result of citrus-agent-maven-plugin

### LINKED ISSUE #1727: Citrus Agent Maven Plugin RunTest Mojo test goal breaks when receiving a test fault json structure
**Citrus Version**

5.0.1

**Expected behavior**

When running tests using the citrus agent maven plugin and receiving a json response for a test fault the execution should report this (displaying a minus in the output) and continue waiting for the overall test execution to finish.

**Actual behavior**

Execution of the plugin breaks, error message is as follows:

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 178, column: 13] (through reference chain: java.lang.Object[][22]->org.citrusframework.agent.plugin.model.RemoteResult["cause"])
```

**Explanation and suggested fix**

RemoteResult class of Citrus Agent Maven Plugin has field `cause` of type `java.lang.String` where the TestResult class of Citrus Agent has field `cause`declared with type `java.lang.Throwable`.

Declaring field `cause` in RemoteResult class also as Throwable fixes this issue. 

I will provide a PR for this.

cquoss:
Closes #1727


bbortt:
hm. is this breaking in any way @christophd?

christophd:
Hi @cquoss, thanks for the contribution and for tracking down the deserialization issue!

The root cause is actually on the **agent server side**, not in the Maven plugin client. During the Jackson 3.x migration (`8d145dc940`), the custom serializer registration in `JsonSupport` was changed from:

```java
// Jackson 2.x — mutates the mapper in place
OBJECT_MAPPER.setSerializerFactory(OBJECT_MAPPER.getSerializerFactory()
        .withAdditionalSerializers(new SimpleSerializers(List.of(new DurationSerializer(), new ThrowableSerializer()))));
```

to:

```java
// Jackson 3.x — withAdditionalSerializers() returns a NEW factory, but the result is discarded
builder.serializerFactory()
        .withAdditionalSerializers(new SimpleSerializers(List.of(new DurationSerializer(), new ThrowableSerializer())));
```

In Jackson 3.x, `withAdditionalSerializers()` returns a new `SerializerFactory` rather than mutating in place — but the returned factory was never assigned back to the builder. This means the `ThrowableSerializer` (which serializes `Throwable` as a simple string `"ClassName - message"`) was silently skipped, causing Jackson to fall back to its default `Throwable` serialization that produces a complex JSON object. That's what triggers the `MismatchedInputException` on the Maven plugin side.

The proper fix is to register the serializers through the `AgentModule` (which is already wired via `.addModule()`). I've opened #1729 with that server-side fix plus a test that verifies the serialization format.

christophd:
Closing in favor of #1729

PR #1729, opened 2026-09-17
fix(#1727): register custom serializers via module in agent JsonSupport

### LINKED ISSUE #1727: Citrus Agent Maven Plugin RunTest Mojo test goal breaks when receiving a test fault json structure

christophd:
Closes #1727

## Summary

The Jackson 3.x migration in `8d145dc940` changed the serializer registration in `JsonSupport` from `OBJECT_MAPPER.setSerializerFactory()` (which mutates in place) to `builder.serializerFactory().withAdditionalSerializers()`. However, in Jackson 3.x `withAdditionalSerializers()` returns a **new** factory — the result was never assigned back to the builder, so `ThrowableSerializer` and `DurationSerializer` were silently skipped.

Without the custom `ThrowableSerializer`, Jackson's default serialization produces a complex JSON object for `Throwable` fields. When the Maven plugin client tries to deserialize this into `RemoteResult.cause` (declared as `String`), it fails with `MismatchedInputException: Cannot deserialize value of type java.lang.String from Object value`.

### Fix

Register both custom serializers via `addSerializer()` in the `AgentModule`, which is already added to the builder via `.addModule()`. This is the idiomatic Jackson approach and avoids the mutable-vs-immutable factory pitfall.

### Test

Added `JsonSupportTest` verifying that:
- `Throwable` is serialized as a simple string (`"ClassName - message"`)
- `Duration` is serialized as a numeric millisecond value
- Successful `TestResult` serializes with expected fields

Generated with [Claude Code](https://claude.com/claude-code)


christophd:
Closes #1728

PR #1738, opened 2026-09-21
feat(#1737): upgrade Playwright to 1.63.0 and expose missing driver capabilities

### LINKED ISSUE #1737: Upgrade Playwright to 1.63.0 and expose missing driver capabilities
The `citrus-playwright` connector pins Playwright Java `1.61.0`, two minor releases behind the current `1.63.0`.

Two classes of gap follow from that: capabilities **added in 1.62/1.63** that the connector cannot reach until the driver moves, and capabilities **already available at 1.61** that were never exposed as Citrus actions.

### Upgrade

`playwright.version` `1.61.0` → `1.63.0` (Chromium 153, Firefox 155, WebKit 26.6).

No removed APIs are referenced — `_react`, `_vue`, `:light`, `devtools`, `page.accessibility()` and `backgroundPage` all return zero hits in the module.

### Capabilities to expose

Blocked on the upgrade:

- Screenshot `type` (png/jpeg/**webp**), `quality` and `fullPage` — today only `setPath` is passed
- `locator.visible()` — the recommended replacement for the `:visible` CSS pseudo-class
- Selector-less `page.frameLocator()` — search any frame in the subtree without naming the iframe
- `locator.waitForFunction()` — wait on an app-specific predicate
- `scroll` option (`auto`/`none`) to opt out of automatic scroll-into-view
- `ariaSnapshots` / `screenSnapshots` on `tracing.start()`
- Origin-scoped list of `httpCredentials`
- `opfs` in `storageState()`
- `page.onDialogClosed()` — assert a dialog was closed

Already available at 1.61, never exposed:

- `page.screencast()` — recording, action annotations, chapter overlays, real-time frames
- `locator.drop()` — synthetic drag-and-drop of files or clipboard data
- `browserContext.credentials()` — WebAuthn virtual authenticator for passkey flows
- `tracing.startHar()` / `stopHar()` — HAR recording as a first-class tracing API
- `assertThat(locator).matchesAriaSnapshot()` — structural aria matching alongside the existing substring check

### Notes

- Everything is additive; no existing public API changes.
- Aria/screen snapshots and screencast recording are expensive and will default to **off**.
- HAR bodies will default to omitted so credentials are not written into the artifact.
- WebAuthn private key material must not reach test variables, logs or failure evidence.

Follows on from #1714.

asuruceanu:
Closes #1737.

Upgrades the `citrus-playwright` driver from `1.61.0` to `1.63.0` and exposes 14 capabilities that were unreachable — some because they are new in 1.62/1.63, others because they shipped in earlier releases and were never surfaced as Citrus actions.

## The upgrade

`playwright.version` `1.61.0` → `1.63.0` (Chromium 153, Firefox 155, WebKit 26.6), committed on its own so it can be reverted with a single line.

No source change was required. The APIs removed in 1.56–1.58 have zero references in the module, and the existing 246 unit tests plus `PlaywrightBrowserIT` passed unchanged against the new driver before any capability work began.

## What is now available

| Capability | Notes |
|---|---|
| Screenshot `type` / `quality` / `fullPage` | png, jpeg, **webp**; format inferred from the extension |
| `LocatorSpec.visible()` | Replaces the `:visible` pseudo-class; a filter, so `.visible().nth(1)` is the second *visible* match |
| Selector-less `frame()` | Searches every frame in the subtree; ambiguity reports the locator and scope |
| `waitFor().function(...)` | Predicate wait on the matched element, with `${var}` support |
| `scroll("none")` | On the seven commands the driver supports it for |
| `tracing().ariaSnapshots()` / `screenSnapshots()` | Off by default — they add a capture per action |
| `tracing().startHar()` / `stopHar()` | Bodies **omitted by default** |
| `storage().opfs(true)` | Origin private file system in storage state |
| `dialog().verifyClosed()` | Backed by `onDialogClosed`; listener removed within the action |
| `verify().ariaSnapshotMatches(...)` | Structural matching, **alongside** the existing substring check |
| `drop()` | Synthetic drag-and-drop of files or clipboard data |
| Origin-scoped `httpCredentials` | Repeatable `@HttpCredential`; explicit `contextOptions` still wins |
| `credentials()` | WebAuthn virtual authenticator for passkey flows |
| `screencast()` | Recording, action annotations, chapter overlays |

## Security decisions

Three surfaces that can carry secrets were closed by design rather than by redacting after the fact:

- **HAR bodies** default to `OMIT`, so the secret never reaches the file.
- **WebAuthn private keys** never reach test variables, logs or failure evidence — asserted by a test that fails if a seeded key appears in the context.
- **The aria snapshot matcher** raises a driver `AssertionError` containing page content, bypassing `SecretPatternRedactor`. It is caught and rethrown as a Citrus `ValidationException` with the configured patterns applied first.

## Decisions worth a reviewer's attention

- **`scroll` is not on the shared `ElementActionBuilder`.** Verified against the jar: `setScroll` exists on six pointer option classes only. Putting it on the shared builder would have offered it on nine actions that cannot honour it, so it lives on the two builders that can, with a build-time guard.
- **`ariaSnapshotMatches` is added, not substituted.** Substring and structural matching are different semantics; replacing the existing check would have silently changed the meaning of tests already written against it.
- **The screencast chapter setter is `chapterDescription`, not `description`.** `AbstractTestActionBuilder` already declares `description(String)` returning the self type, so the obvious name would have been a same-erasure override silently redirecting every `description(...)` call into the chapter overlay. A test now asserts the two stay independent.
- **One existing test assertion changed.** `PlaywrightActionBuilderTest` asserted that a frame action without a selector throws — that encoded the contract this PR deliberately removes. It is replaced with one covering the remaining fail-fast path, not deleted.

## Not in scope

15 actions still have no XML/YAML element (`console`, `context`, `cookies`, `dialog`, `download`, `emulate`, `enter`, `frame`, `network`, `page`, `pageObject`, `pdf`, `permissions`, `storage`, `tracing`). That gap pre-dates this PR; options added here to those actions are therefore Java-DSL only, and the manual now says which. Closing it is follow-up work.

Migrating `NetworkCaptureRegistry` to `page.requests()` was investigated and **rejected**: the registry redacts at capture time, `NetworkRecord` is a snapshot rather than an RPC-backed handle, and failed requests have no equivalent in `requests()`.

## Verification

- **321 unit tests** (from 246), all green
- **7 integration tests** against real Chromium 153, including a new one driving visible-only locators, suppressed scrolling, a predicate wait, a nested-frame click, a drop, a WebP screenshot, and HAR + screencast recording in one run. The screenshot assertion reads the `RIFF`/`WEBP` signature, so a PNG written to a `.webp` path would fail it.
- `./mvnw clean install -pl connectors/citrus-playwright -am` clean, zero new compiler warnings
- Reference manual: 255 → 506 lines, with an example per new action


connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/endpoint/PlaywrightEndpointBuilder.java
```diff
@@ -24,6 +24,10 @@
 import java.util.Arrays;
 import java.util.List;
 
+import java.util.List;
```
christophd: **Duplicate import.** `java.util.List` is already imported at line 25. This will cause a compile warning or error.

```suggestion
```

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/actions/DialogAction.java
```diff
@@ -54,12 +59,51 @@ public DialogAction(Builder builder) {
...
+        closed.set(false);
+        Consumer<Dialog> listener = dialog -> closed.set(true);
+        browser.getCurrentPage().onDialogClosed(listener);
+
+        try {
+            if (triggerScript != null) {
+                browser.getCurrentPage().evaluate(LocatorResolver.resolve(triggerScript, context));
+            }
+        } finally {
+            if (triggerScript != null) {
```
christophd: **Listener leak when `triggerScript` is null.** The `onDialogClosed(listener)` call on line 83 always registers the listener, but this `offDialogClosed` call is guarded by `triggerScript != null`. When `verifyClosed()` is used without a trigger script, the listener is never removed. The Javadoc on the method says "The listener is always removed again so it cannot leak into a following action" — that contract isn't met in the no-script path.

Consider moving the `offDialogClosed` call outside the `if` guard:

```suggestion
            browser.getCurrentPage().offDialogClosed(listener);
```

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/xml/Screencast.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @XmlAttribute
+    public void setTitle(String title) {
+        this.title = title;
+    }
+
+    @XmlAttribute
+    public void setCommand(String command) {
+        switch (command.trim().toLowerCase(java.util.Locale.ENGLISH).replace('_', '-')) {
+            case "start" -> delegate.start(path);
```
christophd, line 55: **Minor inconsistency:** This uses `Locale.ENGLISH` while every other DSL string normalizer in this PR (`ScrollModes`, `HarPolicies`, `ScreenshotTypes`, `Credentials`) uses `Locale.ROOT`. The difference is irrelevant for ASCII case conversion, but the inconsistency may confuse future contributors. The same applies to the YAML `Screencast` counterpart.

```suggestion
        switch (command.trim().toLowerCase(java.util.Locale.ROOT).replace('_', '-')) {
```

christophd (commented):
## PR #1738 Review

This is a well-structured, thoroughly tested PR. The contributor has clearly thought about security (HAR bodies omitted by default, WebAuthn keys not leaked, ARIA snapshot redaction), API design (scroll only on commands that support it, `chapterDescription` vs `description`, additive changes only), and has provided comprehensive test coverage (321 unit tests, 7 integration tests, documentation updates).

Two actionable findings below — the duplicate import and the listener cleanup in `executeVerifyClosed`. The rest are minor observations.

This review is scoped to the project's contribution rules and conventions. It does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud static analysis.

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/endpoint/PlaywrightEndpointBuilder.java
asuruceanu: Fixed in `94c1487d5464f452234052060ea8848ee4afc13b`. Removed the duplicate `java.util.List` and moved `com.microsoft.playwright.options.HttpCredentials` up into the Playwright import block, so the file has one ordered group per package root again.

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/actions/DialogAction.java
asuruceanu: Good catch - fixed in `94c1487d5464f452234052060ea8848ee4afc13b`. The `offDialogClosed` call is now unconditional in the `finally` block, so the listener is removed on every path and the javadoc contract holds. Added `shouldDeregisterClosedListenerWithoutTriggerScript` alongside the existing trigger-script test so the guard cannot come back unnoticed.

connectors/citrus-playwright/src/main/java/org/citrusframework/playwright/xml/Screencast.java
asuruceanu: Fixed in `94c1487d5464f452234052060ea8848ee4afc13b`. Both the XML and YAML `Screencast` wrappers now use `Locale.ROOT`, matching `ScrollModes`, `HarPolicies`, `ScreenshotTypes` and `Credentials`. Behaviour is unchanged for the ASCII command names. Left `VerifyAction`'s `Locale.ENGLISH` alone - that one predates this PR.

PR #1742, opened 2026-09-21
ci(#1741): build only changed modules on pull requests

### LINKED ISSUE #1741: ci: pull request builds rebuild the whole reactor regardless of what changed
### Problem

Every pull request build compiles and tests all 94 modules, even when the change is confined to a single leaf module.

A recent run ([35614418484](https://github.com/citrusframework/citrus/actions/runs/35614418484)) took **37m46s**, of which the Maven build itself was 36m25s. Breakdown by plugin goal:

| goal | time | share |
|---|---|---|
| `failsafe:integration-test` | 973s | 45% |
| `surefire:test` | 940s | 43% |
| `javadoc:jar` | 164s | 7.6% |
| `compiler:compile` + `testCompile` | 51s | 2.3% |

So ~89% of the build is test execution, and most of it is unrelated to any given change.

### Impact

The reactor dependency graph shows how much of this is avoidable. Measuring the blast radius of each module (the module plus everything downstream of it):

- median module blast radius: **3% of the build**
- 58 of 94 modules have a blast radius under 10%
- only 9 modules have a blast radius over 90%

Replaying the last 299 commits on `main` against that graph, restricting the build to changed modules plus their dependents would cut total CI time by **35%**, taking the mean build from 36.4 min to 23.5 min. Roughly 27% of commits would finish in under 4 minutes.

The median does not improve much — 38% of commits touch `citrus-api`, `citrus-base` or `citrus-spring`, whose dependents are effectively the whole reactor — but the gain for contributors working on a single connector or endpoint is large. A change confined to `citrus-playwright`, for example, needs 13 modules instead of 94.

### Proposal

Adopt [gitflow-incremental-builder](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder) for pull request builds only, so the reactor is restricted to modules changed on the branch plus everything downstream of them. Pushes to `main`, release builds and local builds would continue to build the full reactor.

### Notes

- Changes to the root `pom.xml` (including every dependency version bump) correctly fall back to a full build, since any module may be affected.
- Documentation under `src/manual` has to be excluded explicitly, otherwise every doc edit maps onto the root module and pulls in the whole reactor.

asuruceanu:
Closes #1741

Registers [gitflow-incremental-builder](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder) 4.5.1 so a pull request build covers only the modules the branch touched plus everything downstream of them.

**Disabled by default.** Local builds, pushes to `main` and release builds continue to build the full reactor; only the pull request job opts in, with `-Dgib.disable=false`.

### Changes

| file | what |
|---|---|
| `.mvn/extensions.xml` | registers the extension |
| `pom.xml` | GIB configuration, `gib.disable=true` by default |
| `.github/workflows/build.yml` | `fetch-depth: 0`, fetch `main`, enable on `pull_request`, print the selected modules |

### Measurements

Replaying the last 299 commits on `main` against this configuration:

| | today | with this change |
|---|---|---|
| single connector or endpoint pull request | 36.4 min | **~1.3 min** |
| mean pull request build | 36.4 min | **23.5 min** |
| median pull request build | 36.4 min | 34.8 min |
| commits finishing under ~4 min | 0% | **27%** |
| total CI time | 181.4 h | 117.3 h (**-35%**) |

Verified end to end, not only in the reactor listing. A change confined to `citrus-playwright`:

```
13 modules (1 tested + 12 upstream with tests skipped)
6 Playwright integration tests ran, 0 failures
BUILD SUCCESS in 59.7s
```

Selection was checked against the dependency graph for several shapes of change:

| change | modules selected |
|---|---|
| `citrus-playwright` only | 13 |
| `citrus-ftp` only | 31 |
| documentation only | 0 |
| root `pom.xml` | 93 (full) |

### Design notes

**Upstream modules are rebuilt, not restored from cache.** `buildUpstream=true` with `skipTestsForUpstreamModules=true`. Resolving them from the cached `~/.m2` SNAPSHOTs would be faster, but that cache is restored via a prefix key and can hold artifacts from an arbitrary earlier build of `main`.

**`src/manual` is excluded explicitly.** Documentation lives at the repository root, so without `gib.excludePathsMatching` every doc edit maps onto the root module and pulls in the whole reactor. No `.md` or `.adoc` file anywhere under `src/*/resources` is used as a test fixture, so nothing the exclusion skips is exercised by a test.

**Root `pom.xml` changes still build everything**, including every dependency version bump. This is intentional: a `dependencyManagement` edit can affect any module, and there is no safe way to narrow it without moving the version properties out of the root pom.

### Failure modes

- If the reference branch cannot be resolved (for example a broken fetch), Maven exits 1 with an explicit message rather than silently selecting nothing. Checked by pointing `gib.referenceBranch` at a non-existent ref.
- `buildDownstream=true`, so a changed module always drags its dependents in.
- The `Show modules selected by the incremental build` step prints the impacted module list on every pull request run, so selection stays auditable.
- Reverting is a one-line change: drop `-Dgib.disable=false` from the workflow.

### Known limitation

The median build barely moves. 38% of commits touch `citrus-api`, `citrus-base` or `citrus-spring`, whose dependents are effectively the whole reactor, and nothing incremental can help those. The benefit is concentrated on connector and endpoint changes.


.mvn/extensions.xml
```diff
@@ -0,0 +1,30 @@
...
+  ~ limitations under the License.
+  -->
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0"
+            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <!--
+    Restricts the reactor to modules touched by the current branch plus everything
+    downstream of them. Inactive unless -Dgib.enabled=true is passed, so local
+    builds and `./mvnw clean install` keep building the full reactor.
+  -->
```
christophd, line 24: Nit: this comment says `-Dgib.enabled=true` but the actual mechanism used in `pom.xml` and the workflow is `gib.disable=true/false`.

```suggestion
    Restricts the reactor to modules touched by the current branch plus everything
    downstream of them. Inactive unless -Dgib.disable=false is passed, so local
    builds and `./mvnw clean install` keep building the full reactor.
```

christophd (commented):
Nice work — the measurements are thorough and the safety rails (disabled by default, downstream always included, full build on root pom changes) are well thought through. One minor comment inline plus a design suggestion.

This review does not replace specialized review tools such as CodeRabbit, Sourcery, or SonarCloud.

pom.xml
```diff
@@ -154,6 +154,31 @@
...
+    <gib.buildUpstreamMode>changed</gib.buildUpstreamMode>
+    <gib.skipTestsForUpstreamModules>true</gib.skipTestsForUpstreamModules>
+    <!--
+      Documentation lives at the repository root, so without this every doc edit
+      maps onto the root module and drags the whole reactor in.
+    -->
+    <gib.excludePathsMatching>^src/manual/.*|.*\.adoc$|.*\.md$</gib.excludePathsMatching>
+    <gib.buildAllIfNoChanges>false</gib.buildAllIfNoChanges>
+    <gib.logImpactedTo>target/gib-impacted-modules.log</gib.logImpactedTo>
+
```
christophd, line 181: Suggestion: consider wrapping all the GIB configuration properties (except `<gib.disable>true</gib.disable>` which should stay in the default properties as a safety net) inside a Maven profile, e.g. `<profile><id>gib</id>...</profile>`. That way:

1. The default `<properties>` section stays clean — only the one-line disable guard lives there.
2. All GIB tuning (`referenceBranch`, `buildDownstream`, `excludePathsMatching`, etc.) is scoped to when incremental builds are actually active.
3. The workflow becomes `./mvnw ... -Pgib` instead of `-Dgib.disable=false`, which is more self-documenting and idiomatic Maven.

.mvn/extensions.xml
asuruceanu: Good catch, that was just wrong — there was never a `gib.enabled` property. Fixed, and since the profile suggestion below is applied too, the comment now refers to the `gib` profile rather than a system property.

pom.xml
asuruceanu: Applied — agreed on all three points, the workflow reads much better as `-Pgib`.

The default `<properties>` now carries only the guard:

```xml
<gib.disable>true</gib.disable>
```

and everything else moved into `<profile><id>gib</id>`, which flips it to `false` alongside the tuning. The workflow passes `-P${MAVEN_PROFILES}` with `playwright-runtimes,gib` on `pull_request` and `playwright-runtimes` otherwise.

I did not want to assume that a profile-supplied property reaches GIB, since it is a Maven extension reading properties in `afterProjectsRead` rather than a plugin resolving parameters, so I checked it:

| invocation | result |
|---|---|
| `./mvnw validate` | `gitflow-incremental-builder is disabled`, 94 modules |
| `./mvnw validate -Pgib` | extension starts, change detection runs |
| `-Pgib`, change confined to `citrus-ftp` | 31 modules, 7 impacted — identical to the previous `-Dgib.disable=false` behaviour |

The keep-the-guard-in-properties instinct turned out to be the right one for a second reason: it also sets the fail-safe direction. If the profile ever fails to activate, `gib.disable` stays `true` and the build falls back to the full reactor rather than to a partial one.

One interaction worth recording, unrelated to the refactor but surfaced by it. `playwright-runtimes` is declared in `connectors/citrus-playwright/pom.xml` rather than the root pom, so on any pull request where GIB prunes that module Maven logs:

```
[WARNING] The requested profile "playwright-runtimes" could not be activated because it does not exist.
```

It is a warning and the build still exits 0 — I verified the combined `-Pplaywright-runtimes,gib` invocation end to end — but it will appear in the logs of most incremental runs, so it is better known about than puzzled over later.

christophd (approved):
Awesome! Many thanks!

